PR #18, opened 2015-01-04
Update to latest Rust.

JP-Ellis:
Fixes `deriving` being renamed to `derive`, and partial fix for
associated types (still need to wait for certain traits to implement
associated types).


BurntSushi:
Thank you! It seems I've fallen a little behind. I'll bring everything up to date now.

BurntSushi:
FYI, I've pushed `0.12.3` with this fix (and others) to crates.io.

PR #19, opened 2015-01-04
Keep track of Master

JP-Ellis:


BurntSushi:
Thanks!

BurntSushi:
@JP-Ellis I just push an updated `iter-owned` branch. However, I wasn't thinking and did a rebase instead of a merge (and then forced pushed it). So you'll have to deal with that on the receiving side. Sorry! Won't do it again!

(And I haven't had a chance to noodle on the problem yet.)

PR #21, opened 2015-01-23
Fix build for Show/String->Debug/Display.

### LINKED ISSUE #20: method `detail` is not a member of trait `StdError`
Just upgraded my nightly rust to the `2015-1-23` build, and now running into compilation errors on one of my projects:
```
   Compiling csv v0.12.9
/Users/me/.cargo/registry/src/github.com-1ecc6299db9ec823/csv-0.12.9/src/lib.rs:317:5: 317:66 error: method `detail` is not a member of trait `StdError`
/Users/me/.cargo/registry/src/github.com-1ecc6299db9ec823/csv-0.12.9/src/lib.rs:317     fn detail(&self) -> Option<String> { Some(self.to_string()) }
                                                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
error: aborting due to previous error
Could not compile `csv`.
```

This sounds like it's related to https://github.com/rust-lang/rust/commit/3cb9fa26ef9905c00a29ea577fb55a12a91c8e7b . I tried just deleting the `detail` method indicated by the error message (since `cause` was already implemented, and the RFC made it sound like that's what it turned into), but that led to another problem that I didn't understand. So I gave up.

But here's an issue, anyway.

alexchandel:
Fixes #20.

### DIFF
--- a/src/bench.rs
+++ b/src/bench.rs
@@ -1,4 +1,4 @@
-use std::fmt::Show;
+use std::fmt::{Debug, Display};
 use std::io::{self, ByRefReader};
 use std::io::Reader as IoReader;
 use stdtest::Bencher;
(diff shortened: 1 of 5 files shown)


BurntSushi:
Excellent, thank you! `0.12.10` should be available on crates.io now.

PR #25, opened 2015-02-03
Make decoder work with newest nightly build.

timhabermaas:
`FromStr::from_str` now returns `Result` not `Option`.

I'm not sure if we want to do something useful with the error wrapped in `Err`. I currently ignore it.


timhabermaas:
Oh, looks like I missed the examples. Will fix them as well.

BurntSushi:
Awesome, thank you!

BurntSushi:
I've uploaded this fix to crates.io in `0.12.14`.

timhabermaas:
Nice! :)

PR #38, opened 2016-03-26
Add Writer::to_string and Writer::to_bytes methods

Xion:
Docs on `Writer::from_memory` mention the existence of `to_string` and `to_bytes` methods that can be used to retrieve the encoded data at any time. Unfortunately, those methods don't actually exist.

This PR adds them. Both consume the `Writer<Vec<u8>>` and produce either a `String` or the underlying `Vec<u8>` without copying.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -117,7 +117,7 @@ impl<W: io::Write> Writer<W> {
 
 impl Writer<Vec<u8>> {
     /// Creates a new CSV writer that writes to an in memory buffer. At any
-    /// time, `to_string` or `to_bytes` can be called to retrieve the
+    /// time, `as_string` or `as_bytes` can be called to retrieve the
     /// cumulative CSV data.
     pub fn from_memory() -> Writer<Vec<u8>> {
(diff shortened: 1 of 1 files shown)


src/writer.rs
```diff
@@ -143,6 +143,24 @@ impl Writer<Vec<u8>> {
             Ok(()) => &**self.buf.get_ref(),
         }
     }
+
+    /// Returns the written CSV data as owned string.
+    pub fn to_string(self) -> String {
```
BurntSushi: Could you rename this to `into_string`? (Per conventions in `std`.) Similarly for `to_bytes` -> `into_bytes`.

BurntSushi:
The documentation for `from_memory` should say `as_bytes`/`as_string`, not `to_bytes`/`to_string`.

Xion:
Fixed both.

BurntSushi:
Great, thank you!

BurntSushi:
This PR is now on crates.io in `0.14.4`.

PR #40, opened 2016-04-21
make the last example run

kindlychung:
Solved this error:

```
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Parse(LocatableError { record: 5, field: 1, err: UnequalLengths { expected: 3, got: 1 } })', ../src/libcore/result.rs:785
note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: Process didn't exit successfully: `target/debug/stridist` (exit code: 101)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,28 +175,27 @@ use std::io::{self, Write};
 use csv::index::{Indexed, create_index};
 
 fn main() {
-  let data = "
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks!

PR #43, opened 2016-06-04
Serde support

adwhit:
This PR adds `serde` support to rust-csv. For now it is best considered experimental. It is strictly opt-in and should not affect the API apart from changing the traits bounds on Reader::decode (Decodable -> Deserialize) and Writer::encode (Encodable -> Serialize).

It is a fairly major refactor so please review and comment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,21 @@
 language: rust
+rust:
+  - "nightly"
+  - "beta"
+env:
+  - FEATURES="serde" TEST_FEATURES="serde-test"
+  - FEATURES="rustc-serialize" TEST_FEATURES="rustc-serialize"
 matrix:
-  include:
-    - rust: "nightly"
-      env: TEST_SUITE=suite_nightly
-    - rust: "beta"
-      env: TEST_SUITE=suite_beta
+  exclude:
+    - rust: beta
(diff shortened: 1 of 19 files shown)


adwhit:
Closed as it has bitrotted too much to be useful now. Hope to have another crack at some point

BurntSushi:
@adwhit Ah right, thanks for working on this though! Note that I am in the process of doing a complete rewrite of this crate (you can monitor progress on the `rewrite` branch), and the plan is to ditch `rustc-serialize` and just use Serde. (There will also be support for `#![no_std]`.)

adwhit:
Oh that's good to hear, I look forward to it. Thanks for making a really useful crate.

BurntSushi:
@adwhit No problem. Sorry for dropping the ball on this PR. I remember when it came in, but it just slipped too far down in my inbox and it never recovered. :-(

dtolnay:
@BurntSushi [here is how I find those](https://github.com/search?p=7&q=is%3Apr+is%3Aopen+user%3Aburntsushi&type=Issues&utf8=%E2%9C%93) :wink:

BurntSushi:
@dtolnay Oh. My. God. I've never looked at that before. That's horrifying.

PR #53, opened 2016-10-12
Use `extend_from_slice` instead of iterator `extend`

fhartwig:
Extending a Vec using `extend_from_slice` is much faster than extending from an iterator. This reduces the time needed for quoting a string by 40-65% (40% on 50-char strings, 65% on 1000-char strings) if the char contains no escape characters. Also, I think it looks nicer 😃

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ doc = false
 [dependencies]
 byteorder = "0.5"
 rustc-serialize = "0.3"
+memchr = "0.1.11"
 
 [dev-dependencies]
 regex = "0.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -183,6 +183,7 @@
 #![deny(missing_docs)]
 
 extern crate byteorder;
+extern crate memchr;
 extern crate rustc_serialize;
 
 use std::error::Error as StdError;
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -3,6 +3,7 @@ use std::io::{self, Write};
 use std::path::Path;
 use std::str;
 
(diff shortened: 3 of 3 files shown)


fhartwig:
Incidentally, replacing the `s.iter().position(|&v| v == self.quote)` in the same method by `memchr(self.quote, s)` reduces the runtime by an order of magnitude. I wasn't sure if you consider quoting performance to be worth the extra dependency, though. I'd be happy to make that change as well if you think it's worth it.

BurntSushi:
@fhartwig This is good work, and I'd probably accept the PR, but this is undoubtedly a breaking change since this probably increases the minimum Rust version required to compile `csv`. I guess I'm not opposed to merging this and doing a semver incompatible bump, since `extend_from_slice` has been around since 1.6.

> Incidentally, replacing the s.iter().position(|&v| v == self.quote) in the same method by memchr(self.quote, s) reduces the runtime by an order of magnitude. I wasn't sure if you consider quoting performance to be worth the extra dependency, though. I'd be happy to make that change as well if you think it's worth it.

This sounds like a good idea and I'd definitely be happy to add `memchr` as a dependency. If it's convenient for you, adding it into this PR is fine.

fhartwig:
I've pushed another commit for the `memchr` change. I only made this PR because I thought it was a no-downsides performance improvement, but I hadn't considered compatibility with older compilers, so feel free to close this if you prefer.

BurntSushi:
Awesome, thank you!

Backwards compatibility is important, but it's been a while since we had a version bump, so I'm fine with doing that.

Issue #54, opened 2016-11-22
Request: multiline records optional


workanator:
Hello,

could you make multiline records optional? I'm asking because the files I process sometimes contain broken records where a field has open quote but has not closing quote and because of that the parser continues to searching for the end of the field on the next text line in the file. That conforms to CSV standard I know but it could be awesome if you can make that behavior optional.

Changes I suggest are
1. Add the new function to `Reader`, e.g. `fn multiline(self, yes: bool) -> Reader<R>`
2. By default multiline is set to `true` and that is what the current framework do.
3. If multiline is set to `false` and the end of the field is not found in the current line `Reader` returns an error.

BurntSushi:
@workanator Does this have precedent in other CSV parsers? What happens with Python's in particular?

Note that there may be performance implications here.

This crate does do its best to parse broken CSV (even if it isn't spec conforming), but there's a limit to what we can do.

workanator:
@BurntSushi Sorry I do not work on Python. I migrate my CSV tool set from Perl to Rust. I used the module [Text::CSV_XS](http://search.cpan.org/search?query=text%3A%3Acsv_xs&mode=all) it does not have such option but it parsing behavior differs from your framework. If that is not possible it's ok. I will look for a workaround solution.

BurntSushi:
@workanator I think it would be fine to experiment with a feature like this and see if it can be done reasonably simply (and without sacrificing performance), but I'm unlikely to do it soon.

(I'd like to revamp the API in this crate, but just haven't had the chance to do so.)

workanator:
@BurntSushi I will try to add that feature. If I success I will create a PR for review. Thanks.

BurntSushi:
This issue is inactive, and it's not clear to me that we should actually support this.

ark-:
I'd be interested in this feature. For context as to why: Microsoft Excel exports multi line CSV files. I commonly use this to export data from excel conveniently and often find myself having to manually remove newlines found between double quotes `"` in order for Rust libraries to parse this. Obviously this is lossy in that if I ever wanted to put the data back into Excel then the newlines are lost. Not a big deal but just adding to the dialog.

BurntSushi:
[...]

PR #55, opened 2016-11-24
Multiline Feature

workanator:
Hello @BurntSushi,

I implement the request #54 as a crate feature. So by default the feature is disabled and the crate works without any change and when the feature the behavoir of `Reader` can be changed with `multiline` option. Enabling the feature may slighly reduce crate performance.

In short terms the behavior of `Reader` is:

- `multiline`=`true` - No changes.
- `multiline`=`false` - Depends on `flexible` option:
  * `flexible`=`true` - Return record where the number of fields may differ from the first record.
  * `flexible`=`false` - Return error `ParseError::UnclosedQuotedField`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,9 @@ test = false
 bench = false
 doc = false
 
+[features]
+multiline = []
+
 [dependencies]
 byteorder = "0.5"
 rustc-serialize = "0.3"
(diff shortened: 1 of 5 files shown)


BurntSushi:
This feature doesn't exist, but this PR is against an older version of the CSV crate. More broadly, I don't think I'd accept a PR for this by adding new Cargo features, which greatly increases maintenance burden.

I'm happy to revisit this, but I'm still skeptical of whether it belongs in this crate or not.

PR #57, opened 2017-01-23
Fix example

aheart:
The break concluded the loop after the first row so the count was never greater than 1.


BurntSushi:
No longer relevant after the rewrite (#71). Thanks though!

Issue #59, opened 2017-02-07
Writer: Implement a way to obtain the inner writer


nicotoumetis:
I'm using a `csv::Writer<flate2::GzEncoder<File>>`, and I'm having issues with writes not being synced to disk properly. I suspect I need to call `File::sync_all()`. However, AFAIK there is currently no way to access the inner Write object `W` from a `csv::Writer<W>`.

Proposal: implement a method on `Writer` similar to `BufWriter`'s `into_inner` or `GzEncoder`'s `finish()`. Something like:

```rust
fn into_inner (self) -> Result<W> {
    self.buf.into_inner()
};
```

PR #60, opened 2017-02-07
Add into_inner method to Writer

nicoburns:
Adds an `into_inner` method to `Writer` to allow consumers of the library to regain access to the underlying Write instance. See #59


BurntSushi:
This was done in #71 (a complete rewrite).

PR #66, opened 2017-04-07
Don't break on EndOfRecord

mschmo:
Otherwise it will hit `break;` on the first occurrence of `EndOfRecord`, which will always result with a max count of 1.


BurntSushi:
Thanks! The failure in CI is interesting... but it looks unrelated.

PR #79, opened 2017-06-10
Eliminate "illegal for this to emit an output byte without consuming the input byte" possibility by using an enum

EvinRobertson:
I was reading your code to see how the DFA transformation worked. The pair of booleans returned from transition_nfa seemed like a "smell," so I changed it to an enum. Tests pass and benchmark results are unchanged (this initialization code isn't the hot-path).

Change transition_nfa to return an NfaInputAction enum instead of a pair of booleans. Only 3 of the 4 possibilities were allowed, enforced by documentation and an assert. Now perhaps slightly clearer what each state transition is doing.

### DIFF
--- a/csv-core/src/reader.rs
+++ b/csv-core/src/reader.rs
@@ -933,9 +933,6 @@ impl Reader {
     /// This returns the next NFA state along with an NfaInputAction that
     /// indicates what should be done with the input byte (nothing for an epsilon
     /// transition, copied to a caller provided output buffer, or discarded).
-    ///
-    /// Note that this is more like a DFA than a true NFA; you cannot transition
-    /// to multiple states.
     #[inline(always)]
     fn transition_nfa(
         &self,
--- a/csv-core/src/reader.rs
+++ b/csv-core/src/reader.rs
@@ -382,6 +382,19 @@ pub enum ReadRecordNoCopyResult {
     End,
 }
 
+/// What should be done with input bytes during an NFA transition
+#[derive(Clone, Debug, Eq, PartialEq)]
+enum NfaInputAction {
+    // Do not consume an input byte
+    Epsilon,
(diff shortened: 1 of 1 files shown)


csv-core/src/reader.rs
```diff
@@ -913,106 +926,105 @@ impl Reader {
...
-    /// buffer. For example, seeing a field delimiter will consume an input
-    /// byte but should not copy it to the output buffer.
+    /// This returns the next NFA state along with an NfaInputAction that
+    /// indicates what should be done with the input byte (nothing for an epsilon
+    /// transition, copied to a caller provided output buffer, or discarded).
     ///
-    /// It is illegal for this to emit an output byte without consuming the
-    /// input byte.
+    /// Note that this is more like a DFA than a true NFA; you cannot transition
+    /// to multiple states.
```
BurntSushi: I disagree with this statement. The presence of epsilon transitions puts this firmly in NFA territory.

EvinRobertson: I agree that DFAs do not have epsilon transitions, and supporting them here reduces some redundancy in transition_nfa. But I was expecting to see something dealing with the power set of possible states. There doesn't seem to be any nondeterminism here, which is what I was trying to remark on.

BurntSushi: > But I was expecting to see something dealing with the power set of possible states.

Indeed. It's there, in `build_dfa`, but is probably not what you're expecting. Since the alphabet and the set of states is so small, it simply runs the NFA through every possible set of inputs. The epsilon closure is computed by virtue of running the NFA for each input byte until it reaches a non-epsilon transition. For example, the DFA state `StartRecord` includes the following NFA states: `StartRecord`, `StartField`, `EndFieldTerm`, `InRecordTerm`. The implementation somewhat obscures this.

Issue #93, opened 2017-10-10
Support (de)serialization of tuple of structs with header row


jturner314:
The new `serde` functionality is convenient for reading/writing CSV files with a header row, but only when a CSV record corresponds to a single struct in your program. However, a common case is to read/write a CSV record that corresponds to two or more structs. For example, I might have

```rust
struct Input {
    x: f64,
    y: f64,
}

struct Properties {
    prop1: f64,
    prop2: f64,
}
```

and want to get a CSV file such as:

```csv
x,y,prop1,prop2
1,2,3,4
```

by serializing an instance of `(Input, Properties)`.

With `csv 1.0.0-beta.4`, to achieve this and ensure the correct field names, I have to create a new struct `InputWithProperties` that combines the fields from the two structs and write methods to convert between `(Input, Properties)` and `InputWithProperties`.

It would be much more convenient of the `csv` crate supported serialization/deserialization of tuples of structs. I would suggest restricting this functionality to tuples of structs with unique field names, although that wouldn't be strictly necessary if the ordering in the tuple was used.

I may get around to implementing this and submitting a PR at some point, but I did want to suggest this feature before 1.0.0 is stabilized in case implementing it would require breaking changes.

By the way, thanks for creating the `csv` crate!

BurntSushi:
@jturner314 Thanks for the thoughtful feature request!

I suppose in principle I'm not opposed to this, but I'm not sure how hard this is to do (or even if it's possible at all).

jturner314:
@BurntSushi You're welcome! I did [implement this feature](https://github.com/jturner314/typed_csv) for `csv 0.15` using `rustc-serialize`. However, I haven't worked with `serde` enough yet to know if it's possible to do this with `serde`.

BurntSushi:
@jturner314 My experience with Serde is that I've been able to find a way to do anything with rustc-serialize and then some. So the fact that you did it with rustc-serialize seems promising!

jturner314:
@BurntSushi I spent some time working on the implementation of this feature for `Writer::serialize()`. I have a few of questions:

1) Currently, whether or not headers get written depends not only on the value of `has_headers` but also on whether the type being serialized has field names (instead of just returning an error when `has_headers == true` and the type doesn't have field names). I appreciate the convenience of this approach from the user's perspective, but on first impression, it seems inconsistent and surprising (and it's more complex to implement). Do you want to keep this behavior?

[...]

PR #94, opened 2017-10-13
Implement serialize for collections of structs

jturner314:
This PR implements the serialization part of #93. It ended up being more lines changed than I hoped, but a large portion of the changes are new tests and updated documentation. This PR should preserve exsting behavior in the public API, with the exception that some types that used to give an error when called with `serialize` now work.

As part of this PR, I spit off the header serialization functionality into a separate `SeHeader` type. Although this added quite a few lines of code, it made the implementation conceptually simpler. If you'd prefer, I can recombine `SeRecord` and `SeHeader`.

Please let me know if I should make any changes.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -739,14 +739,20 @@ impl<W: io::Write> Writer<W> {
     ///
     /// # Rules
     ///
-    /// For the most part, any Rust type that maps straight-forwardly to a CSV
-    /// record is supported. This includes structs, tuples and tuple structs.
-    /// Other Rust types, such as `Vec`s, arrays, maps and enums have a more
-    /// complicated story. In general, when working with CSV data, one should
-    /// avoid *nested sequences* as much as possible.
+    /// The behavior of `serialize` is fairly simple:
+    ///
+    /// 1. Nested containers (tuples, `Vec`s, structs, etc.) are always
+    ///    flattened (depth-first order).
+    ///
+    /// 2. If `has_headers` is `true` and the type contains field names, then
+    ///    a header row is automatically generated.
+    ///
+    /// However, some container types cannot be serialized, and if
+    /// `has_headers` is `true`, there are some additional restrictions on the
+    /// types that can be serialized. See below for details.
     ///
     /// For the purpose of this section, Rust types can be divided into three
-    /// categories: scalars, non-struct collections, and structs.
+    /// categories: scalars, non-struct containers, and structs.
     ///
     /// ## Scalars
     ///
@@ -816,9 +822,9 @@ impl<W: io::Write> Writer<W> {
     /// }
     /// ```
     ///
-    /// ## Non-Struct Collections
+    /// ## Non-Struct Containers
     ///
-    /// Nested collections are flattened to their scalar components, with the
+    /// Nested containers are flattened to their scalar components, with the
     /// exeption of a few types that are not allowed:
     ///
     /// | Name | Example Type | Example Value | Output |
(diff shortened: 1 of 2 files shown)


src/writer.rs
```diff
@@ -745,123 +745,166 @@ impl<W: io::Write> Writer<W> {
...
-                        HeaderState::DidNotWrite
-                    };
+    ///
+    /// However, if `has_headers` were enabled in the above example, then
+    /// serialization would return an error. Speficially, when `has_headers` is
+    /// `true`, there are two restrictions:
+    ///
+    /// 1. Named field values in structs must be scalars.
+    ///
+    /// 2. All scalars must be named field values in structs.
```
BurntSushi: Does this get formatted as a list? You might need to delete the empty line before (2).

src/writer.rs
```diff
@@ -745,123 +745,166 @@ impl<W: io::Write> Writer<W> {
...
-                        HeaderState::DidWrite
-                    } else {
-                        HeaderState::DidNotWrite
-                    };
+    ///
+    /// However, if `has_headers` were enabled in the above example, then
+    /// serialization would return an error. Speficially, when `has_headers` is
+    /// `true`, there are two restrictions:
+    ///
+    /// 1. Named field values in structs must be scalars.
```
BurntSushi: I think the whole point of this PR is that named field values can be other structs, right?

BurntSushi:
The tables you added are phenomenal by the way. Great idea!

jturner314:
> I am super worried about the complexity this is adding (not just to the implementation but to the contract of serialize as well), but I'm not sure what to do about that.

Hmm, I understand the concern.

From the user's perspective, the behavior is fairly simple: containers (including structs) are always flattened, and if `has_headers` is `true` and the type contains field names, headers are automatically generated. The complexity added by this PR is in the two new restrictions in what the user is *not* allowed to serialize when `has_headers` is `true`, and they'll get an error in those cases. I pushed some reworded documentation. Does it help?

From the implementation perspective, the complexity is in the state machine represented by `serializer::HeaderState`, primarily due to the uncertainty in whether headers actually need to be written. I pushed a commit adding a diagram for the state machine. Does that help? If not, it might be possible to simplify the state machine somewhat by serializing the headers in two passes: (1) determine whether headers need to be written (i.e. determine if the type has any field names) and (2) actually write the headers (while verifying the type meets the restrictions).

> I think the whole point of this PR is that named field values can be other structs, right?

Not quite; this PR makes it so that structs (with named fields) can be inside tuples, `Vec`s, etc. However, 
 tuples, `Vec`s, etc. still cannot be inside named field values. The reason for this restriction is to make sure that all scalars have a corresponding field name. For example

```rust
(
    Foo { x: 5, y: true },
    Bar { z: 3.14 },
)
```

is fine because all of the scalar values (`5`, `true`, and `3.14`) have corresponding field names (`x`, `y`, and `z`). In contrast,

```rust
Foo { x: 5, y: (true, false) },
```

is prohibited because the value of `y` is a container, and it's unclear how to handle this case. (Should the field name `y` be duplicated for each value in the tuple; should all values of the tuple be placed in one field and quoted; etc.)

I tried to clarify the restrictions with the examples in the table; maybe I need to come up with a better way to explain them?

> Does this get formatted as a list? You might need to delete the empty line before (2).

Yes, it gets formatted as a list. The empty line ensures that the Markdown parser generates `<p></p>` tags around each item in the list (which I think looks a little nicer).

jturner314:
> Let me know when you think this is ready to merge!

I pushed one more change to the diagram. As soon as the CI tests pass, I think it's ready to merge.

> Also, are you planning on doing the deserialization side of this as well?

I'd like to, but it may be a while before I get around to it.

Thanks for working with me on this PR!

BurntSushi:
All righty, I've pushed `1.0.0-beta.5` out to crates.io with this change! Thanks again!

PR #95, opened 2017-10-13
Fix alignment of examples in docstrings

jturner314:
I came across this issue when updating the docs for `Writer::serialize` in #94, but I just noticed that it appears elsewhere too. This PR is necessary for the Markdown headings to render correctly, e.g. the `Example` heading for [`csv::invalid_option`](https://docs.rs/csv/1.0.0-beta.5/csv/fn.invalid_option.html).

Edit: Without this PR, there's also an extra space before all of the other lines in the code examples.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -331,9 +331,9 @@ impl Default for Terminator {
 /// # fn main() { example().unwrap(); }
 /// fn example() -> Result<(), Box<Error>> {
 ///     let data = "\
-///a,b,c
-///5,\"\",xyz
-///";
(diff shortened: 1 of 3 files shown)


BurntSushi:
Oh man, *this* is what was causing the headers  not to render correctly?!? I was wondering about that and just figured it was some weird Markdown bug. Nice find!

jturner314:
You're welcome! By the way, I just noticed that you're also the author of ripgrep. Many thanks! I use rg all the time, including for this PR. :)

PR #96, opened 2017-11-04
Updated link for uspop-null.csv

anna-liao:
Dead link for uspop-null.csv.  Updated with correct link.


BurntSushi:
@anna-liao Awesome, thank you. :-)

Issue #78, opened 2017-05-29
Leading spaces in header fields


vmx:
I came across a [CSV file](https://archive.org/details/nycTaxiTripData2013) which had leading spaces in the header fields:

    medallion, hack_license, vendor_id, pickup_datetime, payment_type, fare_amount, surcharge, mta_tax, tip_amount, tolls_amount, total_amount

When I deserialized it with Serde, each field has a space in front. Would it make sense to have an option to trim the header fields or to transform them in a more general way?

BurntSushi:
If your struct is like this:

```rust
#[derive(Deserialize)]
struct Record {
  medallion: String,
  hack_license: String,
  ...
}
```

Then you could do:

```rust
#[derive(Deserialize)]
struct Record {
  #[serde(rename = " medallion")]
  medallion: String,
  #[serde(rename = " medallion")]
  hack_license: String,
  ...
}
```

I'm not sure whether this is common enough to elevate this to a more convenient feature. Certainly, I've seen other CSV parsers have options like, "trim all whitespace around field values," which is something that could be feasibly added with some minor additional cost.

vmx:
I'm doing a CSV to JSON conversion, hence I ended up with:

```rust
struct Fare {
    medallion: String,
    #[serde(rename(deserialize = " hack_license", serialize = "hack_license"))]
    hack_license: String,
    ...
}
```

Which isn't that nice.

BurntSushi:
Here is what I'd suggest we do. In the `csv` crate, we should add a new `Trim` enum type and a new `trim` method on [`ReaderBuilder`](https://docs.rs/csv/1.0.0-beta.3/csv/struct.ReaderBuilder.html) that accepts a value with type `Trim`. The `Trim` enum should be defined as follows:

```rust
pub enum Trim {
    None,
    Headers,
    /// Hints that destructuring should not be exhaustive.
    ///
    /// This enum may grow additional variants, so this makes sure clients
    /// don't count on exhaustive matching. (Otherwise, adding a new variant
    /// could break existing code.)
    #[doc(hidden)]
    __Nonexhaustive,
}
```

`Trim::None` should be the default, which matches existing behavior. The `Headers` option should correspond to trimming only header values. We can add other variants later that permit trimming values as well, but it seems easiest to start with headers and does solve the OP's problem.

[...]

PR #97, opened 2017-11-13
Accept whitespace trimming settings

medwards:
Fixes burntsushi/rust-csv#78

Per my comment in #78, I'm popping whitespace and updating the record bounds to reflect the changes.

Some thoughts:
* I don't know if match or if statements are more idiomatic in this case. Let me know if you have an opinion.
* If the user turns off headers then I apply trimming to the first row if `Trim::All` or `Trim::Fields` are set. I think this is reasonable but I could imagine that maybe `Trim::Headers` should be `Trim::FirstRow` or something in which case it isn't reasonable.
* Just now I was thinking maybe `Trim::Fields` should be `Trim::Records` instead.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -5,14 +5,14 @@ use std::path::Path;
 use std::result;
 
 use csv_core::{
-    Reader as CoreReader, ReaderBuilder as CoreReaderBuilder,
+    Reader as CoreReader, ReaderBuilder as CoreReaderBuilder
 };
 use serde::de::DeserializeOwned;
 
 use byte_record::{self, ByteRecord, Position};
 use error::{ErrorKind, Result, Utf8Error, new_error};
 use string_record::{self, StringRecord};
-use Terminator;
+use {Terminator, Trim};
 
 /// Builds a CSV reader with various configuration knobs.
 ///
@@ -24,6 +24,7 @@ pub struct ReaderBuilder {
     capacity: usize,
     flexible: bool,
     has_headers: bool,
+    trim: Trim,
     /// The underlying CSV parser builder.
     ///
     /// We explicitly put this on the heap because CoreReaderBuilder embeds an
@@ -38,6 +39,7 @@ impl Default for ReaderBuilder {
             capacity: 8 * (1<<10),
             flexible: false,
             has_headers: true,
+            trim: Trim::default(),
             builder: Box::new(CoreReaderBuilder::default()),
         }
     }
@@ -316,6 +318,12 @@ impl ReaderBuilder {
     /// ```
     pub fn flexible(&mut self, yes: bool) -> &mut ReaderBuilder {
         self.flexible = yes;
+        self
+    }
+
+    /// Set whitespace preservation behaviour
+    pub fn trim(&mut self, trim: Trim) -> &mut ReaderBuilder {
+        self.trim = trim;
         self
     }
 
@@ -710,6 +718,7 @@ struct ReaderState {
     /// set, every record must have the same number of fields, or else an error
     /// is reported.
     flexible: bool,
+    trim: Trim,
     /// The number of fields in the first record parsed.
     first_field_count: Option<u64>,
     /// The current position of the parser.
@@ -776,6 +785,7 @@ impl<R: io::Read> Reader<R> {
                 headers: None,
                 has_headers: builder.has_headers,
                 flexible: builder.flexible,
+                trim: builder.trim,
                 first_field_count: None,
                 cur_pos: Position::new(),
                 first: false,
@@ -1447,18 +1457,34 @@ impl<R: io::Read> Reader<R> {
     ) {
         // If we have string headers, then get byte headers. But if we have
         // byte headers, then get the string headers (or a UTF-8 error).
-        let (str_headers, byte_headers) = match headers {
+        let (mut str_headers, mut byte_headers) = match headers {
             Ok(string) => {
                 let bytes = string.clone().into_byte_record();
                 (Ok(string), bytes)
             }
             Err(bytes) => {
                 match StringRecord::from_byte_record(bytes.clone()) {
-                    Ok(str_headers) => (Ok(str_headers), bytes),
-                    Err(err) => (Err(err.utf8_error().clone()), bytes),
+                    Ok(str_headers) => {
+                        (Ok(str_headers), bytes)
+                    },
+                    Err(err) => {
+                        (Err(err.utf8_error().clone()), bytes)
+                    },
                 }
             }
         };
+
+        if Trim::should_trim_headers(self.state.trim) {
+            let _ = str_headers.as_mut().map(|record| record.trim());
+            byte_headers = match str_headers.as_ref() {
+                Ok(record) => record.clone().into_byte_record(),
+                Err(_) => {
+                    byte_headers.trim();
+                    byte_headers
+                }
+            }
+        }
+
         self.state.headers = Some(Headers {
             byte_record: byte_headers,
             string_record: str_headers,
@@ -1505,7 +1531,11 @@ impl<R: io::Read> Reader<R> {
     /// }
     /// ```
     pub fn read_record(&mut self, record: &mut StringRecord) -> Result<bool> {
-        string_record::read(self, record)
+        let result = string_record::read(self, record);
+        if Trim::should_trim_fields(self.state.trim) {
+            record.trim();
+        }
+        result
     }
 
     /// Read a single row into the given byte record. Returns false when no
@@ -1558,6 +1588,9 @@ impl<R: io::Read> Reader<R> {
             if let Some(ref headers) = self.state.headers {
                 self.state.first = true;
                 record.clone_from(&headers.byte_record);
+                if Trim::should_trim_fields(self.state.trim) {
+                    record.trim();
+                }
                 return Ok(!record.is_empty());
             }
         }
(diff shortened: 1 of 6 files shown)


medwards:
Added some updates, its much cleaner and a bit faster.

I've also done some performance analysis now. Here is master:
`test count_nfl_read_bytes ... bench: 1,969,508 ns/iter (+/- 97,802) = 692 MB/s`
And if you add trimming you get
`test count_nfl_read_bytes ... bench: 3,838,782 ns/iter (+/- 601,554) = 355 MB/s`
([full bench results](https://gist.github.com/medwards/3604b4a2b4f1f130bef44690160690f2))

The average performance impact across all benchmarks is roughly 46% slower (the error bars might be up to 10% off that, not helped by weird things like write benchmarks being mysteriously slower too).

medwards:
@BurntSushi: Hey I know you're probably busy with impl period stuff, but I was hoping you could take a preliminary look at this and at least give me an acceptable/unacceptable in principle. I'm happy to come up with an alternative approach, but I'll have really spotty internet in a week so if its necessary I'd like to figure out the high-level stuff this week.

src/reader.rs
```diff
@@ -1449,13 +1455,21 @@ impl<R: io::Read> Reader<R> {
         // byte headers, then get the string headers (or a UTF-8 error).
         let (str_headers, byte_headers) = match headers {
             Ok(string) => {
-                let bytes = string.clone().into_byte_record();
+                let mut bytes = string.clone().into_byte_record();
+                match self.core.trim {
```
BurntSushi: I think it would be better to define (unexported) predicates on the `Trim` type. e.g., `should_trim_headers` and `should_trim_field`.

src/reader.rs
```diff
@@ -1449,13 +1455,21 @@ impl<R: io::Read> Reader<R> {
         // byte headers, then get the string headers (or a UTF-8 error).
         let (str_headers, byte_headers) = match headers {
             Ok(string) => {
-                let bytes = string.clone().into_byte_record();
+                let mut bytes = string.clone().into_byte_record();
```
BurntSushi: I don't think `string` gets trimmed here..?

src/reader.rs
```diff
@@ -1449,13 +1455,21 @@ impl<R: io::Read> Reader<R> {
...
+                    CoreTrim::Headers | CoreTrim::All => bytes.trim(),
+                    _ => ()
+                }
                 (Ok(string), bytes)
             }
             Err(bytes) => {
-                match StringRecord::from_byte_record(bytes.clone()) {
-                    Ok(str_headers) => (Ok(str_headers), bytes),
-                    Err(err) => (Err(err.utf8_error().clone()), bytes),
+                let mut final_bytes = bytes;
```
BurntSushi: Can you use `Err(mut bytes)` instead?

src/reader.rs
```diff
@@ -1449,13 +1455,21 @@ impl<R: io::Read> Reader<R> {
...
+                if self.core.trim == CoreTrim::Headers || self.core.trim == CoreTrim::All {
```
BurntSushi: Use a predicate here.

src/reader.rs
```diff
@@ -1558,6 +1572,9 @@ impl<R: io::Read> Reader<R> {
             if let Some(ref headers) = self.state.headers {
                 self.state.first = true;
                 record.clone_from(&headers.byte_record);
+                if self.core.trim == CoreTrim::Fields || self.core.trim == CoreTrim::All {
```
BurntSushi: Use a predicate here.

src/reader.rs
```diff
@@ -1569,8 +1586,14 @@ impl<R: io::Read> Reader<R> {
             // never return the first row. Instead, we should attempt to
             // read and return the next one.
             if self.state.has_headers {
-                return self.read_byte_record_impl(record);
+                let ret_val = self.read_byte_record_impl(record);
```
BurntSushi: Call this `let result = ...` please.

src/reader.rs
```diff
@@ -1569,8 +1586,14 @@ impl<R: io::Read> Reader<R> {
...
+                if self.core.trim == CoreTrim::Fields || self.core.trim == CoreTrim::All {
```
BurntSushi: Use a predicate here.

src/reader.rs
```diff
@@ -1569,8 +1586,14 @@ impl<R: io::Read> Reader<R> {
...
+                    record.trim();
+                }
+                return ret_val;
             }
+        } else if self.core.trim == CoreTrim::Fields || self.core.trim == CoreTrim::All {
```
BurntSushi: Use a predicate here.

src/reader.rs
```diff
@@ -2168,6 +2191,67 @@ mod tests {
         assert!(!rdr.read_byte_record(&mut rec).unwrap());
     }
 
+    #[test]
+    fn read_trimmed_records() {
```
BurntSushi: Please split this test up.

BurntSushi:
> re: just trimming ASCII whitespace - aha I misread your comment and thought you were limiting the scope to just ByteRecord trimming.

Ah yeah sorry about the confusion. I think the issue here is that we're introducing a `trim` option on the reader itself, and if we're going to go that route, then we need to go all the way and implement proper Unicode trimming for `StringRecord` as well.

There does exist an incremental step here, should you wish to take it. You could get rid of the `trim` option on the reader and instead just focus on providing the `trim` method on `ByteRecord`. We could then incrementally add `trim` to `StringRecord`, and then finally add it to the reader. It's up to you though!

> re: core_trim - at first I did it because I thought I would be making the change in core-csv, then after that turned out not to be true it felt kind of like polluting the non-core reader to add it (all of the knobs currently seem to be in core-csv). I'm fine either way and it should be easy

Yup I definitely don't see this as pollution. It just so happens that most of the knobs exist at the csv core level, but not all knobs do! For example, the flexible and header knobs exist at the `csv` crate level.

> re: more tests for ByteRecord.trim - are the doctests insufficient? Several (all?) of the cases you list are covered by the doctest and I got the sense that several functions in ByteRecord had their core usecases tested through doctests and just weird edge cases went into unit tests. lmk either way.

Oh hmm I missed those. They don't cover all the cases though. The one I don't see is a test for the empty record. It might also be good to add more variations with 1-field record vs multi-field records.

medwards:
> There does exist an incremental step here, should you wish to take it

Naw, I want to be able to use this so I'll take it to the finish line and make sure its exposed.

> I don't see is a test for the empty record. It might also be good to add more variations with 1-field record vs multi-field records.

K, I'll try to think of some single field edge cases. Empty record is literally just an empty string? With multiple fields?

BurntSushi:
> Empty record is literally just an empty string? With multiple fields?

Either, I suppose. The goal here should be to test `trim` on any valid record value.

medwards:
Did this on the plane. Needs a lot more tests but I am curious to see whether the recursive approach in StringRecord.trim could be used in ByteRecord and what the performance difference is.

Some thoughts I'd like comments on now:
* trim setting is in ReaderState with similar settings, but it feels weird there. Just less weird than on Reader which is pretty bare
* I think I got set_headers_impl hashed out. I think its really hard to understand what its doing and the Result that contains two potential input values doesn't help. Is Result here being used as a bootstrap for an "either/or" struct?
** note that if the utf8 error in the Result<StringRecord> has an index then that index may be invalid after trimmming
* Every StringRecord that gets trimmed first gets trimmed as a ByteRecord as part of string_record::read so there is some duplicated and wasted efffort. Is this OK for now?

unit tests and benchmarks coming...

medwards:
@BurntSushi I think I'm ready for another review.

There are some notes in my previous comment, as to the new stuff:
* Tell me if the unit tests are comprehensive enough for you. I could have gotten into single record "trim left/trim right/ignore internal whitespace" tests but I felt this was excessive given the doctests. lmk either way
* only added benchmarks to one of the sample sets, lemme know if where you want it or for everyone
```
test count_mbta_iter_bytes                 ... bench:   1,980,153 ns/iter (+/- 14,208) = 365 MB/s
test count_mbta_iter_bytes_trimmed         ... bench:   3,484,228 ns/iter (+/- 23,231) = 207 MB/s
test count_mbta_iter_str                   ... bench:   2,527,966 ns/iter (+/- 9,659) = 286 MB/s
test count_mbta_iter_str_trimmed           ... bench:   7,438,755 ns/iter (+/- 112,220) = 97 MB/s
```

medwards:
I did the experiment I mentioned above:
The current `ByteRecord.trim` implementation is about 36% faster than a recursive version and isn't much less readable. I've left the current version in the PR.

medwards:
@BurntSushi Hey man, hope you had a swell New Years :) Is there anything here left for me to do?

BurntSushi:
@medwards I wanted to try to get this merged, so I took this over the finish line. I hope you don't mind. You can see my PR that is based off your work in #106.

Many of my changes were stylistic, but there were a few logical changes as well:

* When reading a header, byte records should only be trimmed via ASCII whitespace rules. I believe this PR would check to see if there was a string record, trim that, and then set the byte record to that result.
* Recursion in `StringRecord`'s implementation of `trim` just is not viable. I dare say that recursion anywhere in this library is not viable. :-) I modified it to use a simple for-loop.
* I added a few more tests.
* I added more documentation, particularly on the CSV reader builder, including an example.

medwards:
Hrm, yeah I feel you wanting to push it over the finish line as I felt the same way. Regarding the headers at least its probably best if you fiddle with it. Looking at my old diff I'm not even sure what I was doing in that section of code. At minimum please refer to my comment above regarding `set_headers_impl` readability needing improvement.

PR #119, opened 2018-06-22
I added a less restrictive write_record method ...

hervold:
Currently, it's up to the user to convert values into &[u8], so I've added a method to automate that, like so:

  let rec: &[&Display] = &[&1, &2.2, &"abc"];
  wtr.write_disp_record(rec);

I should add a write_disp_field method as well, but I thought I'd run the idea by you first.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -1,3 +1,4 @@
+use std::fmt::Display;
 use std::fs::File;
 use std::io;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


BurntSushi:
This does not seem like a good fit for this library. It seems straight-forward to just use `write_record`. If you're this operation frequently, then define a helper function. I think adding this to this library would just make the API more confusing for most folks without any commensurate benefit.

PR #128, opened 2018-08-30
chmod -x benches/bench.rs

cuviper:
The mode was changed in commit cddabebe63f5, probably because an editor
was trying to be smart with the first line that looks like a shebang.


BurntSushi:
Nice catch! Thanks!

PR #134, opened 2018-09-25
exclude unneeded files from crates.io

ignatenkobrain:


BurntSushi:
This PR is in `csv 1.0.2` on crates.io.

PR #150, opened 2019-04-04
Attempt to add supprt for i128 and u128

ian-hamlin:
Issue 149.

In the infer_deserialize functions I ended up adding try_positive_integer_64 and try_negative_integer_64, this was to keep a couple of existing unit tests happy, but I'm not 100% sure if that was a good idea, any feedback on that would be welcome.


src/deserializer.rs
```diff
@@ -297,10 +301,14 @@ impl<'r> DeRecord<'r> for DeByteRecord<'r> {
...
-        } else if let Some(n) = try_positive_integer_bytes(x) {
+        } else if let Some(n) = try_positive_integer_bytes_64(x) {
             visitor.visit_u64(n)
-        } else if let Some(n) = try_negative_integer_bytes(x) {
+        } else if let Some(n) = try_negative_integer_bytes_64(x) {
             visitor.visit_i64(n)
+        }else if let Some(n) = try_positive_integer_bytes(x) {
+            visitor.visit_u128(n)
+        } else if let Some(n) = try_negative_integer_bytes(x) {
+            visitor.visit_i128(n)
```
BurntSushi: If you _replace_ the 64-bit parsing with 128-bit parsing completely, do the tests still pass?

If not, then this is fine I think!

src/serializer.rs
```diff
@@ -1085,6 +1119,18 @@ mod tests {
...
+    #[test]
+    fn struct_no_headers_128() {
+        #[derive(Serialize)]
+        struct Foo {
+            x: i128,
+            y: u128,
+        }
+
+        let got = serialize(Foo { x: i128::max_value(), y: u128::max_value() });
+        assert_eq!(got, "170141183460469231731687303715884105727,340282366920938463463374607431768211455\n");
```
BurntSushi: Could you try to make sure code is wrapped to 79 columns inclusive? Thanks!

ian-hamlin: I've wrapped these lines.

src/writer.rs
```diff
@@ -1436,6 +1436,22 @@ mod tests {
...
+        struct Row {
+            foo: i128,
+            bar: f64,
+            baz: bool,
+        }
+
+        let mut wtr = WriterBuilder::new()
+            .has_headers(false)
+            .from_writer(vec![]);
+        wtr.serialize(Row { foo: 9_223_372_036_854_775_808, bar: 42.5, baz: true }).unwrap();
```
BurntSushi: This line still looks longer than 79 columns.

src/deserializer.rs
ian-hamlin: It does break two existing tests. I'm trying to work out why to see if I can remove the 64-bit parsing fully.

src/writer.rs
ian-hamlin: I missed that one, sorry. Should be fixed now.

src/deserializer.rs
BurntSushi: Figured. I don't think you can get rid of. Serde will "cascade" the 8/16/32 bit integers up to u64 automatically, which is why we don't need the smaller sizes. But 128-bit integers are a conditional feature since they are somewhat new, so there's no cascading from 64 bit integers up to 128-bit integers. I figured this was the case, but wasn't 100%.

I think this pretty much looks good to go. I'm going to hold off on merging it until I'm ready to release `csv 1.1.0` though. (Because this increases the MSRV.)

PR #152, opened 2019-04-11
Deserialize hexidecimal numbers

mwilbur:
I recently had the need to deserialize hexidecimal numbers in a CSV file.  I could not find any crate that fit my needs.  This pull request add this ability.

Caveat: I'm new-ish to rust and crazy new to serde and macros.

### DIFF
--- a/src/deserializer.rs
+++ b/src/deserializer.rs
@@ -318,12 +318,12 @@ macro_rules! deserialize_int {
             visitor: V,
         ) -> Result<V::Value, Self::Error> {
             let field = self.next_field()?; 
-            let num = if field.starts_with("0x") {
-                <$inttype>::from_str_radix( &field[2..], 16 )
-            } else {
-                field.parse()
-            };
-
+            let num = 
+                if field.starts_with("0x") {
+                    <$inttype>::from_str_radix( &field[2..], 16 )
+                } else {
+                    field.parse()
+                };
             visitor.$visit( num.map_err(|err| self.error(DEK::ParseInt(err)))?)
         }
     }
--- a/src/deserializer.rs
+++ b/src/deserializer.rs
@@ -312,14 +312,19 @@ impl<'r> DeRecord<'r> for DeByteRecord<'r> {
 }
 
 macro_rules! deserialize_int {
-    ($method:ident, $visit:ident) => {
+    ($method:ident, $visit:ident, $inttype:ty) => {
         fn $method<V: Visitor<'de>>(
             self,
             visitor: V,
         ) -> Result<V::Value, Self::Error> {
-            visitor.$visit(
-                self.next_field()?
-                    .parse().map_err(|err| self.error(DEK::ParseInt(err)))?)
+            let field = self.next_field()?; 
+            let num = if field.starts_with("0x") {
+                <$inttype>::from_str_radix( &field[2..], 16 )
+            } else {
+                field.parse()
+            };
+
+            visitor.$visit( num.map_err(|err| self.error(DEK::ParseInt(err)))?)
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/deserializer.rs
```diff
@@ -312,14 +312,19 @@ impl<'r> DeRecord<'r> for DeByteRecord<'r> {
...
         ) -> Result<V::Value, Self::Error> {
-            visitor.$visit(
-                self.next_field()?
-                    .parse().map_err(|err| self.error(DEK::ParseInt(err)))?)
+            let field = self.next_field()?; 
+            let num = if field.starts_with("0x") {
+                <$inttype>::from_str_radix( &field[2..], 16 )
+            } else {
+                field.parse()
+            };
```
BurntSushi: Could you write this like so:

```rust
let num =
    if field ... {
        ...
    } else {
        ...
    };
```

Thanks!

mwilbur: I thought about using the `deserilize_with` attribute but that would leave me deserializing the whole vector instead of the elements and I wanted to leave that part to the csv deserializer since it seemed redundant to something the library already did well.  I will make the change.

BurntSushi: Just to follow up and clarify here: I don't know what you mean by deserializing the whole vector, but you can use `deserialize_with` on individual fields. See the [`csv::invalid_option` examples](https://docs.rs/csv/1.0.6/csv/fn.invalid_option.html).

mwilbur: Right - I didn't give very good context.  What I want to do is deserialize records like

`WriteBlock,0x7E,0x7D,0x10,0x32,0x75`

to 

```
enum Command {
    WriteBlock,
    ...
}

struct I2cTransaction {
  command: Command,
  address: u8,
  data: Vec<u8>
}
```

My brief foray into deserializtion made me thing that if I used `deserilize_with` on the data field I'd have to take care of the parsing the commas, etc.  However, based on stuff I read later in the day I'm not as sure about that (having read about `deserialize_seq` for example).  But I'm still kind of struggling to keep my head above water with the serde stuff.  I'm good with trying to that road if you're able to give me some pointers :).

BurntSushi: Oh I see, gotya. Yeah in that case I think you'd have to do this:

```rust
struct I2cTransaction {
    ...
    data: Vec<Byte>,
}

struct Byte {
    #[serde(deserialize_with = "...")]
    byte: u8,
}
```

Which is indeed a little annoying.

I don't mean to suggest that you do this, since I do think I agree with this change. Just wanted to throw this out there for full understanding and in case you needed a work-around before I get to merging this and doing a release.

And yeah, serde can be tricky to wrap your head around. I'd say you're doing great so far. :-)

mwilbur: :)   I just pushed the change to address the styling.  My work as a ridiculous set of port filtering rules.

src/deserializer.rs
```diff
@@ -312,14 +312,19 @@ impl<'r> DeRecord<'r> for DeByteRecord<'r> {
...
-                self.next_field()?
-                    .parse().map_err(|err| self.error(DEK::ParseInt(err)))?)
+            let field = self.next_field()?; 
+            let num = 
+                if field.starts_with("0x") {
+                    <$inttype>::from_str_radix( &field[2..], 16 )
+                } else {
+                    field.parse()
+                };
+            visitor.$visit( num.map_err(|err| self.error(DEK::ParseInt(err)))?)
```
BurntSushi: Sorry I missed this before, but please remove the spaces after an opening paren and before a closing one (as above). In general, it's a good idea to follow a style that's consistent with the surrounding code (which does not have these spaces).

Thanks!

mwilbur: My apologies.  I haven't developed Rust style sensitivites yet :).

BurntSushi:
This PR is in `csv 1.0.7` on crates.io!

mwilbur:
Sweet!  My first rust contribution!

PR #174, opened 2019-10-06
Only flush underlying writer on explicit flush call, not full buffer

### LINKED ISSUE #173: Do not flush underlying writer when buffer becomes full

master: `fcdbea357739348b9eed2a0239e180213be78af9`

As mentioned in [this issue of `xsv`](https://github.com/BurntSushi/xsv/issues/151#issuecomment-538790858), it seems that rust-csv is flushing the underlying writer when its internal buffer becomes full. This causes issues for `xsv` as the underlying writer is a [`TabWriter`](https://github.com/BurntSushi/tabwriter) and calling `flush` causes side-effects other than writing out the buffer (specifically, it resets the internal state, "breaking" alignment of subsequently written data)

The reproduction with `xsv` is attached on the linked issue - using a small `rust-csv` buffer size leads to broken tabular layout.

#### What is the expected or desired behavior of the code above?

When the internal buffer becomes full, its contents should be written to the underlying writer, but the underlying writer should not be `flush`ed. This seems to match with other buffering IO libraries, such as [`BufferedWriter` in Java](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/io/BufferedWriter.java#L119-L124) and [rust's stdlib](https://github.com/rust-lang/rust/blob/master/src/libstd/io/buffered.rs#L490)

owst:
Fixes #173

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -1043,7 +1043,7 @@ impl<W: io::Write> Writer<W> {
                     self.state.fields_written += 1;
                     return Ok(());
                 }
-                WriteResult::OutputFull => self.flush()?,
+                WriteResult::OutputFull => self.flush_buf()?,
             }
         }
     }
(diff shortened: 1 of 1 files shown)


owst:
Hi @BurntSushi, any chance you would be able to take a look at this PR? The issue it causes in xsv (https://github.com/BurntSushi/xsv/issues/151) is quite annoying - it'd be great to have it fixed. Thanks!

owst:
Hi @BurntSushi sorry to pester again, but any chance you can take a look at this PR and then hopefully get https://github.com/BurntSushi/xsv/issues/151 resolved?

src/writer.rs
```diff
@@ -1306,6 +1315,47 @@ mod tests {
...
+        wtr.write_byte_record(&ByteRecord::from(vec!["c", "d"])).unwrap();
+        wtr.flush().unwrap();
+        wtr.write_byte_record(&ByteRecord::from(vec!["e", "f"])).unwrap();
+
+        let got = wtr.into_inner().unwrap().to_str();
+
+        // As the buffer size is 4 we should write each record separately, and
+        // flush when explicitly called and implictly in into_inner.
+        assert_eq!(got, ">a,b\n<>c,d\n<!>e,f\n<!");
+    }
```
BurntSushi: Great test, thank you!

BurntSushi:
This PR is in `csv 1.1.2` on crates.io.

PR #175, opened 2019-10-09
refactor: make serde an optional dependency

### LINKED ISSUE #126: Make serde an optional dependency
From the example I have the impression that serde is not actually needed in order to use this crate. If I am correct, then it would be better to specify serde as an optional dependency and compile it only if the user wants it. It would be like this:

`Cargo.toml`
```toml
[dependencies]
serde = {version = "1.0.55", optional = true}

[features]
default = ["serde"]
```

This way, if the user does not want serde, all what it has to do is to put this in his `Cargo.toml`

```toml
[dependencies.csv]
version = "1"
default-features = false
```

lojewalo:
See title! Closes #126.

My concerns:

1. `Cargo.toml` took a beating with required features.
2. Doc comments with serde are ugly.

I think 2 is fine, but 1 feels awful.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,12 +23,15 @@ members = ["csv-core", "csv-index"]
 [lib]
 bench = false
 
+[features]
+default = ["serde"]
+
 [dependencies]
 bstr = { version = "0.2.1", features = ["serde1"] }
 csv-core = { path = "csv-core", version = "0.1.6" }
 itoa = "0.4.3"
 ryu = "1"
-serde = "1.0.55"
+serde = { version = "1.0.55", optional = true }
 
 [dev-dependencies]
 serde = { version = "1.0.55", features = ["derive"] }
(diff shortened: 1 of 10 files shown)


jkcclemens:
I think the rustfmt output looks much worse in this case, but I added a commit just for the sake of CI passing.

Kerollmops:
I just tried you branch and removed all the `[[xxx]] required-features = ["serde"]` configurations lines and it worked without complaining. Do you think those are necessary?

jkcclemens:
Running `cargo test --no-default-features` failed without it when I tried. Did you attempt that?

Kerollmops:
Hum... You are right :) Just tried.

jbaum98:
What's the status of this PR? I would like this feature, so if there's anything I can do to get this merged let me know.

devashishdxt:
@BurntSushi Are there any plans to merge this PR?

BurntSushi:
> @BurntSushi Are there any plans to merge this PR?

Perhaps eventually.

devashishdxt:
> > @BurntSushi Are there any plans to merge this PR?
> 
> Perhaps eventually.

Are there any changes required before merging this?

Issue #176, opened 2019-10-18
Patterns for incrementally parsing chunked data


golddranks:
#### What version of the `csv` crate are you using?

name = "csv"
version = "1.1.1"

#### Briefly describe the question, bug or feature request.

Hi, this is not a bug request, but just a question about the usage patterns.
I'm dealing with streaming IO, and I've got my CSV parsing logic in a function
that is repeatedly called with new chunks of data. I've encountered the following problems:

1 )Repeatedly constructing new Readers from the input slices "resets" the parsing: all info about headers etc. is lost.
2) On exhausting the end of the input slice, I'd rather revert parsing to last complete record and return the unparsed input to be `chained` as a reader with the next chunk instead of trying to parse a record and possibly missing the end of the record.

So my question is: are the current API's intended to accomodate for this usage case, and I'm just failing to find the correct knobs and patterns or I'm I trying to use the csv crate for something it wasn't originally meant for? Would it be possible to add APIs or documentation that made this easier?

#### Include a complete program demonstrating a problem.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b5a60f9ffcaff611186a6d9b4dafc310

#### What is the observed behavior of the code above?

```
Parse chunk: Chain { t: [], u: [99, 111, 108, 95, 97, 44, 99, 111, 108, 95, 98, 44, 99, 111, 108, 95, 99, 10, 48, 97, 97, 97, 97, 44, 48, 98, 98, 98, 98, 44, 48, 99, 99, 99, 99, 10, 49, 97, 97, 97, 97, 44, 49, 98, 98, 98, 98, 44, 49, 99, 99] }
Got record: ByteRecord(["0aaaa", "0bbbb", "0cccc"])
Got record: ByteRecord(["1aaaa", "1bbbb", "1cc"])
Done. Unparsed: []
Parse chunk: Chain { t: [], u: [99, 99, 10] }
Done. Unparsed: []
Parse chunk: Chain { t: [], u: [50, 97, 97, 97, 97, 44, 50, 98, 98, 98, 98] }
Done. Unparsed: []
Parse chunk: Chain { t: [], u: [44, 50, 99, 99, 99, 99, 10] }
Done. Unparsed: []
[...]

PR #177, opened 2019-10-18
Implement complete_only

golddranks:
I sketched a bit with the `complete_only` API idea I presented here: https://github.com/BurntSushi/rust-csv/issues/176#issuecomment-543547885. What do you think?

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -22,6 +22,7 @@ pub struct ReaderBuilder {
     capacity: usize,
     flexible: bool,
     has_headers: bool,
+    complete_only: bool,
     trim: Trim,
     /// The underlying CSV parser builder.
     ///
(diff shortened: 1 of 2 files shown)


golddranks:
I'm also trying to implement the case for getting a mutable reference to the internal reader, but the required infrastructure for incremental parsing isn't there: The `BufReader` is `consume`d every time bytes are read, but in the case of not being able to parse a method, there is no place to dump the unparsed bytes. I was hoping that in the case a complete record isn't parsed, it would be possible not to call `consume` and keep the bytes in `BufReader`, but not consuming bytes seems to prevent it from loading more, which is required for a complete record parse to be found.

golddranks:
It would be possible to store the incremental state in the output buffer, but since there is no user-facing APIs that would continue from where it left, implementing incremental parsing would be a big undertaking...

PR #178, opened 2019-10-18
inner_mut for Reader

golddranks:
More sketching around https://github.com/BurntSushi/rust-csv/issues/176#issuecomment-543547885

I found out that once `only_complete` is there, a method to get a mutable reference to the inner reader _will_ be helpful: it allows one to implement a feeder pattern demonstrated by `test_chunks_2` of `tests/test_chunks.rs`. This will allow incremental parsing without having to scrap and rebuild `Reader` for every chunk.

With these changes, incremental parsing would be, if not easy, at least a lot easier than before.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -22,6 +22,7 @@ pub struct ReaderBuilder {
     capacity: usize,
     flexible: bool,
     has_headers: bool,
+    complete_only: bool,
     trim: Trim,
     /// The underlying CSV parser builder.
     ///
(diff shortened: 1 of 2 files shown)


golddranks:
All right, I experimented with this design with my code base. The verdict: it doesn't work _too_ well because of lifetimes. The purpose of this API was to be able to smuggle data chunk, i.e. a reference to a slice inside the `csv::Reader` during the CSV parsing. However, as `csv::Reader` has the inner reader as its generic parameter, the lifetime on the inner reader "taints" `csv::Reader`. I was trying to have a `csv::Reader` that outlives the loop where the chunks come in one by one, but once the lifetime of the chunk taints `csv::Reader`, that isn't possible anymore.

PR #186, opened 2020-01-17
Remove deprecated std::error::Error::description and use CARGO_MANIFEST_DIR to find the examples/data/ for test cases

WildCryptoFox:
Was looking for unsafe code to eliminate and deprecated warnings to resolve. So far I've only done the latter. Please reconsider how you use unsafe code. Yes, this appears safe.. but it'd be nicer if it were simply safe.

Fortunately for me, this is only a transitive dependency via `[dev-dependencies.criterion]` and does not involve user controlled data.

Additionally, as I use a custom `RUST_TARGET_PATH`, I caught the directory traversal hack you're using and replaced it with `env!("CARGO_MANIFEST_DIR")`.

### DIFF
--- a/src/deserializer.rs
+++ b/src/deserializer.rs
@@ -738,6 +738,7 @@ impl DeserializeError {
 }
 
 impl DeserializeErrorKind {
+    #[allow(deprecated)]
     fn description(&self) -> &str {
         use self::DeserializeErrorKind::*;
 
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -358,12 +358,7 @@ fn example_bin_dir() -> PathBuf {
 
 /// Return the repo root directory path.
 fn repo_dir() -> PathBuf {
-    debug_dir()
-        .parent()
-        .expect("target directory")
-        .parent()
-        .expect("repo directory")
-        .to_path_buf()
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
 }
 
 /// Return the directory containing the example data.


src/deserializer.rs
```diff
@@ -691,11 +691,7 @@ impl SerdeError for DeserializeError {
     }
 }
 
-impl StdError for DeserializeError {
-    fn description(&self) -> &str {
-        self.kind.description()
-    }
```
BurntSushi: This is a breaking change. Please add it back and use `#![allow(deprecated)]`.

WildCryptoFox: I incorrectly assumed std would just call the Display impl... But it can't due to the type. That's nasty.

Is it really a breaking change if the only way to hit a break is to have code like `match error.description() { "hardcoded error strings here" => do something }`? Anyone who should be doing this kind of matching would use `DeserializeErrorKind`'s variants and gain exhaustive coverage.

src/deserializer.rs
```diff
@@ -738,17 +734,17 @@ impl DeserializeError {
...
+            Message(_) => &"deserialization error",
+            Unsupported(_) => &"unsupported deserializer method",
+            UnexpectedEndOfRow => &"expected field, but got end of row",
+            InvalidUtf8(ref err) => err,
+            ParseBool(ref err) => err,
+            ParseInt(ref err) => err,
+            ParseFloat(ref err) => err,
         }
     }
 }
```
WildCryptoFox: I removed this in an edit... not sure how it returned. The Display impl calls this function only on the `UnexpectedEndOfRow` variant, so that may as well skip the indirection.

WildCryptoFox:
> Thanks! I appreciate the intent, but we can't just remove public APIs and releasing csv 2 just for this isn't worth it.

No public API is removed. The [`std::error::Error::description`] carries a provided implementation returning a constant string telling the consumer the function is deprecated. As mentioned in the comment it'll take an obscure consumer to hit a break and such user should fix their code.

Are you sure you want to keep it with `#[allow(deprecated)]`?

> I'm happy to reconsider use of unsafe, but unsolicited generic commentary on the use of unsafe is not appreciated. If you have specific concrete ideas on how to remove unsafe, then I'd of course be very happy to entertain them. Otherwise, I've already considered whether use of unsafe was worth it or not. So without any new information or ideas, I don't see any reason to re-evaluate that.

I do have an idea.. I forgot to mention it - whoops. Okay. The `ByteRecord` currently has no type parameters and contains a `Vec<u8>` as a field. This could be promoted to a type parameter as a default, which `StringRecord` can replace with `String`. The overlapping code can be generic over all types which implement `AsRef<[u8]>` / `AsMut<[u8]>` as necessary.

src/deserializer.rs
BurntSushi: The breaking change is that someone may be using `description` to output a non-empty string that describes the error. With this PR, they'd get an empty string. It's a change in behavior.

It's quite likely that in practice this wouldn't break anyone. But:

1. I don't have the tools or resources to confirm that.
2. The cost of adding `#[allow(deprecated)]` is so small (virtually zero) that it's not worth worrying about.

WildCryptoFox: Actually, this isn't where the deprecation warning comes from; but to get the strs to implement it, we'd need to call the `.description` methods on the other errors; which is deprecated.

WildCryptoFox: They don't get an empty string; they get std's own message `"description() is deprecated; use Display"`. But okay, I'll revert it.

WildCryptoFox:
@BurntSushi Cool. If you're good to take it from here I have other crates to review. If it works, please add `#![forbid(unsafe_code)]` to signal it up to `cargo geiger` consumers.

If you can't get it, I'll try it later.

BurntSushi:
I experimented with removing `unsafe` for a bit, but it looks like it might not be so easy. The issue comes when we want to convert a `ByteRecord` to a `StringRecord` with as little cost as possible. That is, the only cost that should be paid is a UTF-8 check. There are two issues with this. First, if our types look like this:

```rust
pub struct StringRecord(Box<ByteRecordInner<String>>);
pub struct ByteRecord(Box<ByteRecordInner<Vec<u8>>>);
```

then I don't see a way to convert between them while reusing the `Box` allocation without using `unsafe`.

The second issue is that, to minimize allocation, fields are stored in a flat `Vec<u8>` adjacent to one another. This means that the entire record, when adjacent to one another, may be valid UTF-8 _even when each individual field_ may not be. Therefore, UTF-8 validation must be done on each field individually. (With some fast paths for ASCII only records.) This means that the UTF-8 validation cannot be done with safe standard library APIs. Instead, each field must be validated and then there must be an `unsafe` check to the final conversion to `String`.

So it would be possible to get down to two uses of `unsafe`, although saving the `Box` allocation seems a trickier kind of `unsafe` than what the library has right now.

I'll try to work this comment into an implementation comment so that future readers will better understand why avoiding `unsafe` is hard.

Happy to hear other ideas if you have them.

WildCryptoFox:
Hmm. You could do the checked `str::from_utf8` and panic if the impossible case happens. Then check if the compiler is smart enough to optimize it away. If not, weigh the performance cost.

BurntSushi:
Yes. The entire design is rooted in performance. An extra UTF8 check for each record isn't acceptable unfortunately.

WildCryptoFox:
Measured or assumed?

WildCryptoFox:
@BurntSushi Okay. One more idea. Feature gated safety? If someone needs to use it but doesn't care for the performance, then they can opt-in for it to be slower (to avoid something opting it into unsafety).

WildCryptoFox:
Maybe even two flags. One to allow unsafety (enabled by default) and another to force safety (to tame dependencies).

WildCryptoFox:
Or perhaps better, check if a compile-time `--cfg` (non-feature) is set to allow the end users to request for safe-only code using `RUSTFLAGS+='--cfg safety-please'. This could be standardized within the community.

BurntSushi:
> Measured or assumed?

In this particular case, assumed. But it's a pretty safe assumption given past experience. It's a good idea to add a benchmark, so I filed #187, but it's not a priority for me.

> Feature gated safety? If someone needs to use it but doesn't care for the performance, then they can opt-in for it to be slower (to avoid something opting it into unsafety).

Yes, I've heard this idea brought up a few times in other contexts. I'm generally opposed to this. Because it means testing a distinct code path. I'd rather just get the `unsafe` correct.

WildCryptoFox:
@BurntSushi With a little testing (delayed by caching faults) I'm convinced a global `--cfg reduce_unsafe` marker would be viable for users to opt-in for safety over performance.

I'm working on reducing and eliminating unsafe code; and want to globally enforce `--cap-lints=warn -Funsafe-code` with per-crate exceptions. Similarly forbid running builder scripts and proc-macros without approval by source (and dependency) hashing.

How do you feel about this approach?

BurntSushi:
I'm not sure I see how that's materially different from a Cargo feature? As I said, I do not want to go down the path of supporting two different code paths for everything that involves `unsafe`. Sorry.

WildCryptoFox:
@BurntSushi The difference is this can be a global hint by users (not consuming crates); which crate authors may decide to follow, if they have the safe and unsafe variants and the only reason to decide one vs the other is performance. You may not opt into this, but others may, and this is what good tests and CIs are good with.

BurntSushi:
OK, I was only trying to comment on my own personal willingness to abide to such a thing. I didn't realize you were seeking feedback for the general idea. For that, I'd recommend posting to urlo or irlo.

I think my opinion on the general idea is mostly unchanged. I could be wrong though. For libraries like csv, supporting it wouldn't be _that_ big of a burden. But it's not always possible, or at least, could make the code quite a bit more complex and harder to maintain. So in that light, I'm skeptical of the idea applying more generally. I truthfully feel that effort would be better spent 1) eliminating `unsafe` entirely and, failing that, 2) ensuring that what `unsafe` is used is correct.

WildCryptoFox:
This is intended as (3) if unsafe code cannot be removed due to performance hit and it is correct; then allow users to decide if they care.

urlo or irlo? I've opened an issue on the rust-secure-code-wg `https://github.com/rust-secure-code/wg/issues/35` (avoiding unnecessary back-reference).

BurntSushi:
irlo => https://internals.rust-lang.org/

urlo => https://users.rust-lang.org/

This topic is probably good for urlo.

WildCryptoFox:
@BurntSushi [https://users.rust-lang.org/t/rfc-use-cfg-reduce-unsafe-to-signal-preference-of-safe-code-over-performance/37069](Done.) Thanks for reminding me about this one.

BurntSushi:
No problem. Also, I'm cool with GitHub's backreferences if that means anything. Happy to have issues like this lend context.

WildCryptoFox:
@BurntSushi BTW I've implemented [macros and a test case](https://gitlab.com/sio4/code/reduce_unsafe) using `str::from_utf8` as the example and could be used in your code if you change your mind. Note the `debug_assertions` detail for tests.

PR #192, opened 2020-02-14
Fix documentation of Writer::delimiter

golddranks:
I have a debug assertion about the output of `delimiter` being always 1, and I was surprised when this wasn't the case. After looking at the code, it seems that the documentation itself was wrong; it uses one byte to end the quotation and another byte to insert the delimiter, so the maximum output length is not 1, but 2.

Of course, it would be even better if there were a test case to catch this. I happened to catch this with the following CSV:
```
id,name,date,birthday
4,taro,0000/02/13,2020/02/13
5,hoge,2020/02/13,2020/02/13
9999,taro,2020/02/13,2020/02/13
3,taro,2020/bb/13,2020/02/13
98,piyo,2020/02/13,2020/02/13
-1,taro,2020/02/13,2020/02/13
9999,taro,2020/02/13,2020/02/13
5555555555555555555555555555554567898765435678ghdxk jepyfghbx kjeupyf
ghmb jidhm





a,taro,2020/02/13,2020/02/13
"test,,,,,,,,,ID","TA
RO",0010/99/a,0010/99/a
```

### DIFF
--- a/csv-core/src/writer.rs
+++ b/csv-core/src/writer.rs
@@ -313,7 +313,8 @@ impl Writer {
     /// a field delimiter, then nothing is written to `output`
     /// and `WriteResult::OutputFull` is returned. Otherwise,
     /// `WriteResult::InputEmpty` is returned along with the number of bytes
-    /// written to `output` (which is always `1`).
+    /// written to `output` (which is `1` in case of an unquoted
+    /// field, or `2` in case of an end quote and a field separator).
     pub fn delimiter(
         &mut self,
         mut output: &mut [u8],


BurntSushi:
@golddranks Out of curiosity, if you don't mind me asking, what are you using `csv-core` for?

golddranks:
Thanks for asking! Actually I have a company-internal library called sleek_csv. It's basically a wrapper for arenas that consist of three vectors: data, field ends and record ends. The first two are the exact format that your `csv_core` is using. The point of that library is to have a arenas for chunks of tabular data, used for processing and re-packaging in a streaming fashion. I felt that while your main csv library provided very nice conveniences, it didn't provide the exact API we needed for performant streaming, so I used csv_core just for parsing and implemented the wrapping around that myself.

I'm trying to get sleek_csv open sourced during this year. My biggest challenges, so to speak, using csv_core have been with serde. I actually wanted to have serde support similar to what csv has, but for that, I had to copy and paste big chunks of code from csv's `deserialize.rs`.

Anyway, I'm very thankful for you, if not just for the core parsing library!

Ref: my forays into using csv_core
https://github.com/BurntSushi/rust-csv/pull/178
https://github.com/BurntSushi/rust-csv/pull/177

BurntSushi:
@golddranks Ah thanks. I still don't quite understand what you mean by "streaming," since `csv` is of course capable of parsing a stream without reading all of it into memory. I _think_ you mean incremental, but I confess that I don't quite grok what you mean. :-) I'm happy to wait and find out once `sleek_csv` is released though!

Also, yes, implementing Serde support was quite challenging and it is the most common source of bugs for this crate. Serde's data model just does not map particularly well to CSV's untyped nature. There are still major problems with how it works IMO. There are so many caveats about what works and what doesn't that I can't keep anything straight. It's easily the most hated portion of any CSV code I've written to date.

PR #194, opened 2020-02-23
Use latest link to tutorial in README.md rather than 1.0.0

nelsonjchen:


atouchet:
There is a second link here: https://github.com/BurntSushi/rust-csv/pull/194/files#diff-04c6e90faac2675aa89e2176d2eec7d8R36

Issue #155, opened 2019-05-13
#[serde(flatten)] does not seem to work when serializing


0x6273:
#### What version of the `csv` crate are you using?

```
csv = { git = "https://github.com/BurntSushi/rust-csv" }
serde = { version = "1.0", features = ["derive"] }
```

#### Briefly describe the question, bug or feature request.

Flattening a struct using the #[serde(flatten)] field attribute works as expected when deserializing, but not when serializing.

#### Include a complete program demonstrating a problem.

```rust
use serde::{Serialize, Deserialize};

#[derive(Debug, PartialEq, Serialize, Deserialize)]
struct StructA {
	foo: String,
	#[serde(flatten)]
	struct_b: StructB,
}

#[derive(Debug, PartialEq, Serialize, Deserialize)]
struct StructB {
	bar: String,
	baz: String,
}

fn main() {
	let test = StructA {
		foo: "asdf".to_string(),
		struct_b: StructB {
			bar: "asdf".to_string(),
			baz: "asdf".to_string(),
		},
	};

	let json = r#"{"foo":"asdf","bar":"asdf","baz":"asdf"}"#;
	let csv_ = "foo,bar,baz\nasdf,asdf,asdf\n";

	// json serialization works as expected
	assert_eq!(serde_json::to_string(&test).unwrap(), json);

	// so does deserialization
	let test2: StructA = serde_json::from_str(json).unwrap();
	assert_eq!(test, test2);

	// csv deserialization works
	let test3: StructA = csv::Reader::from_reader(csv_.as_bytes())
		.deserialize().next().unwrap().unwrap();
	assert_eq!(test, test3);

	// csv serialization results in an error:
	// Error(Serialize("serializing maps is not supported, ..."))
	let mut output: Vec<u8> = Vec::new();
	{
		let mut writer = csv::Writer::from_writer(&mut output);
		let result = writer.serialize(test);
		assert!(result.is_ok()); // fails
	}
	assert_eq!(std::str::from_utf8(&output).unwrap(), csv_);
}
```

#### What is the observed behavior of the code above?

```
RUST_BACKTRACE=1 cargo +nightly run
   Compiling csv_flatten_test v0.1.0 (/home/ben/Projects/rust/csv_flatten_test)
    Finished dev [unoptimized + debuginfo] target(s) in 1.12s
     Running `target/debug/csv_flatten_test`
thread 'main' panicked at 'assertion failed: result.is_ok()', src/main.rs:46:3
[...]

Issue #98, opened 2017-11-18
Support serializing of maps


NatureShade:
I want to use this library in order to import/export csv files to/from a key-value database, I use HashMap<String, MyValueType> to store this data. This works fine for deserializing, however I just discovered that it was not supported for serializing, something which feels like an inconsistency in the library. This is a feature that I would appreciate.

BurntSushi:
I don't think this is a matter of will, but rather, a missing specification. If someone wants this feature, then they need to propose how it should work, in detail. A principle question that must be answered is the ordering of columns (which ordering?) and whether or not it should be deterministic (and if so, how).

BurntSushi:
(IMO, I think a valid answer to this question is "the library doesn't support it and callers should convert their data type to something else that is already supported." If you're creating a hash map for every record, then you've already given up on making it fast, so the cost of an extra conversion step is a small helper function.)

NatureShade:
You are probably right, I just realized that this was just the push I needed to actually implement a kind of sorted hashmap so that I can get the csv out in the same format that I inserted it in. However it still seems like an inconsistency with support only for deserializing of maps. I don't think that serialization of at least hashmaps has to be deterministic, when you first choose to use them you are probably no longer interested in ordering.

BurntSushi:
I don't really buy inconsistency arguments just for the sake of consistency, particularly when there are complications. Reading into a hash map and writing from a hash map have fundamentally different concerns when it comes to CSV data.

I also don't really agree about determinism. Non-determinism is often surprising when you hit it, and while it is appropriate for `HashMap` to have arbitrary order, I wouldn't want to propagate that out further.

The reason why I think the status quo is best is because it allows the library to not make a choice about determinism and ordering, and forces the caller to do it.

With that said, here are other thoughts:

1. It seems like `BTreeMap` should be allowed, since the ordering is made clear by the type.
2. It may be worthwhile to attempt to pop up a level and focus on the problem you're trying to solve. If the problem is, "I want a simple way to read CSV data without caring much about its structure, but still retain the ability to inspect it dynamically," (which I'm assuming is what you're doing if you're using a `HashMap`), then there are other approaches here. For example, we might consider a new record type that retains ordering information and exposes a map-like API.

synek317:
I think that serializing maps may be required in order to work with #[serde(flatten)] attribute.

Kampfkarren:
[...]

PR #197, opened 2020-04-15
support for writing a header for nested structs

ilya-epifanov:
This covers some of the #98 and #155 use cases.
It makes #[serde(flatten)] unnecessary for nested structs, as CSV is a flat format anyway.

### DIFF
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -448,11 +448,12 @@ enum HeaderState {
 struct SeHeader<'w, W: 'w + io::Write> {
     wtr: &'w mut Writer<W>,
     state: HeaderState,
+    structs_written: usize,
 }
 
 impl<'w, W: io::Write> SeHeader<'w, W> {
     fn new(wtr: &'w mut Writer<W>) -> Self {
-        SeHeader { wtr: wtr, state: HeaderState::Write }
+        SeHeader { wtr: wtr, state: HeaderState::Write, structs_written: 0usize }
     }
 
     fn wrote_header(&self) -> bool {
@@ -665,10 +666,11 @@ impl<'a, 'w, W: io::Write> Serializer for &'a mut SeHeader<'w, W> {
 
     fn serialize_struct(
         self,
-        name: &'static str,
+        _name: &'static str,
         _len: usize,
     ) -> Result<Self::SerializeStruct, Self::Error> {
-        self.handle_container(name)
+        self.structs_written += 1;
+        Ok(self)
     }
 
     fn serialize_struct_variant(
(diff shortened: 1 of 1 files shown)


ilya-epifanov:
@fmorency, @Kampfkarren, could you please take a look if this covers your use cases?

ilya-epifanov:
In particular, take a look at this test: https://github.com/BurntSushi/rust-csv/pull/197/files#diff-5e07616003d425b526178a965e987ce3R1185

ilya-epifanov:
@BurntSushi could you please take a look?

CGMossa:
I've tried to see if this works with my use case, which is nested structs and wishing to write the headers for them, and it didn't.

jeremias-work:
The patch works for my simple use case, two nested structs. It would help me if the PR got merged.

0xpr03:
Would help here also for flattening stuff which I don't want to patch in one big struct.

Luthaf:
This branch works fine for me for serializing nested structs. Thanks a lot @ilya-epifanov!

amrhassan:
Thank you so much @ilya-epifanov. This however fails when serializing nested structs wrapped in enums.

```rust
use serde::Serialize;

#[derive(Serialize)]
struct Row {
    nested_one: Option<Nested>,
    nested_two: Nested,
}

#[derive(Serialize)]
struct Nested { x: i32, y: i32 }

fn main() {
    let mut writer = csv::Writer::from_writer(std::io::stdout());
    let row = Row { nested_one: Some(Nested { x: 1, y: 2 }), nested_two: Nested { x: 3, y: 4}};
    writer.serialize(row).unwrap();
}
```

This panics with `Error(UnequalLengths { pos: None, expected_len: 3, len: 4 })` after writing:
```
nested_one,x,y
1,2,3,4
```

haixuanTao:
Ok, @BurntSushi, this is what I got.

# Summary

There is 2 main step in writing the CSV: - Writing the header and - Writing the values.

Currently the values are written properly even when they are nested in a struct. On the other hand, the header for a `struct` is written as the field name of the struct and not the nested field of the `struct`.

That's the problem this PR tries to solve and that is why this PR mainly modifies the behaviour of `struct SeHeader`.

# How it is implemented

Basically, this PR hack the `serialize_field` by making the writing of the header field conditional to the type of the field. 
It would behave as before for any type that is not a `struct`, but, if it is a `struct`, it is not going to write it down. The state will move down its nested field and everything will work. 

To do that, the PR has set a new field for the `struct SeHeader`. This new field is called `structs_written` (Line 451, 456-459). It is a counter of the number of struct it has seen since the beginning of the writting of the header. This counter is incremented each time `serialize_struct` is called.

Therefore, if the `structs_written` counter changes when encountering a new field that means the new field is a `struct`(Line 802-805). 


# Problem
This however does not work when the `struct` is encapsulated in an `option` as @amrhassan pointed out. 

The underlying problem is that the hack does not work this time as `serialize_struct` is not called when the struct is in option. Instead `serialize_some` is called. 

Now we could try to patch `serialize_some` as well and increment `structs_written` but i did not find a patch for it, and it will look a bit shaky.

A second problem I faced is that we cannot use the `#[serde(flatten)]`, otherwise the writting will fail with `serializing maps is not supported ...` called in `serialize_map`.
This is kind of counter-intuitive as it is the recommended way to use `serde` serialization.

I think a better way to redo this PR for nesting by implementing the `serialize_map` method with some conditioning on the input and use `serde` `#[serde(flatten)]`.

PR #202, opened 2020-05-22
Implement issue #167

### LINKED ISSUE #167: permit caller to define "nullable" semantics
Currently, the Serde deserializer in this crate hardcodes specific nullable semantics. Namely, only fields with an empty value are considered missing. But, it is fairly common in CSV formats to use other sentinels, such as "null," even for otherwise non-String fields. It should be possible to augment `ReaderBuilder` to permit the caller to provide a predicate that determines whether a field value is missing or not.

Once we have the predicate, it should be as simple as plumbing it through and calling it in [`deserialize_option`](https://github.com/BurntSushi/rust-csv/blob/fcdbea357739348b9eed2a0239e180213be78af9/src/deserializer.rs#L453-L465).

See also https://users.rust-lang.org/t/serde-csv-empty-fields-are-the-string-null/31260 for motivation.

derekdreery:
This PR adds the option to specify a function to mean 'empty field' or
`None`, which currently is always assumed to be `b""`. It stores a
function on the heap, and then calls it with the raw contents of the
field to check for `None`ness.

Issues still to address

 - The name for the setting (I don't like the current name).
 - Validating use of `Fn` and `Rc`. I do this to preserve the behavior of the builder methods: they take the `Builder` by reference and so must continue to have ownership of the function in addition to the `Reader`.
 - Testing
 - More documentation.

This commit serves as a proof of concept and a starting point for discussion.

Potentially closes #167 .


derekdreery:
PS it was me in the urlo question :)

derekdreery:
I'm thinking `custom_none_check` might be an OK name?

derekdreery:
friendly bump.

derekdreery:
Friendly bump.

src/deserializer.rs
```diff
@@ -21,11 +21,13 @@ use self::DeserializeErrorKind as DEK;
 pub fn deserialize_string_record<'de, D: Deserialize<'de>>(
     record: &'de StringRecord,
     headers: Option<&'de StringRecord>,
+    missing_field_check: Option<&'de dyn Fn(&str) -> bool>,
```
Mart-Bogdan, line 24: This looks like a breaking change.  New argument to a public function.

derekdreery: hmm let me have a think

DankDumpster:
Any updates on this?

derekdreery:
Annual ping :slightly_smiling_face:

PR #204, opened 2020-06-06
Implement Clone for ByteRecordIter and StringRecordIter

benesch:
@BurntSushi what do you think about this? I figured since the diff was so small it'd be easier to discuss over a PR than an issue.

Cloneable iterators are useful when writing code that is generic over
iterators that needs to be able to iterate twice.


Ten0 (approved):

BurntSushi (approved):
Thanks!

PR #213, opened 2020-11-02
Avoid running into infinite loops when recovering from Reader errors

### LINKED ISSUE #207: Recovering from CSV read errors leads to infinite loops when caused by IO errors
1.1.3

Recovering from CSV errors leads to infinite loops when caused by IO errors

```rust
use {
    csv::Reader,
    std::io::{self, Read},
};

#[derive(Debug, serde_derive::Deserialize)]
struct Deserialized {
    something: String,
}

fn main() {
    for record_result in Reader::from_reader(FailingRead).deserialize::<Deserialized>() {
        match record_result {
            Ok(deserialized) => println!("Got something: {}", deserialized.something),
            Err(failed_line) => eprintln!("This record is broken :("),
        }
    }
}

struct FailingRead;
impl Read for FailingRead {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        Err(io::Error::new(io::ErrorKind::Other, "reader just broke!"))
    }
}
```

#### What is the observed behavior of the code above?

Infinite loop of "This record is broken :(".

#### What is the expected or desired behavior of the code above?

```
This record is broken :(
```
*program exits*

Ten0:
Resolves #207

When reading and parsing a record fails, it's correct behavior in most cases to keep on trying to read and parse the next record.

However when the Error is caused by failing to read from the underlying `impl Read`, the next read would almost always fail with the exact same error.
In that scenario, reading from the underlying `impl Read` should not be attempted when trying to extract the next record, as this may lead to infinite loops.
Instead, the `Reader` will behave in the same way as if an end-of-file had been reached.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -686,6 +686,10 @@ impl ReaderBuilder {
 ///   [`ReaderBuilder`](struct.ReaderBuilder.html).
 /// * When reading CSV data from a resource (like a file), it is possible for
 ///   reading from the underlying resource to fail. This will return an error.
+///   For subsequent calls to the `Reader` after encountering a such error
+///   (unless `seek` is used), it will behave as if end of file had been reached,
+///   in order to avoid running into infinite loops when still attempting to read
+///   the next record when one has errored.
 /// * When reading CSV data into `String` or `&str` fields (e.g., via a
 ///   [`StringRecord`](struct.StringRecord.html)), UTF-8 is strictly
 ///   enforced. If CSV data is invalid UTF-8, then an error is returned. If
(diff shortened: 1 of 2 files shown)


BurntSushi (changes requested):
Awesome work! I love the comments and the way you implemented this. Great job.

One small nit: could you please make sure all lines are wrapped to 79 columns (inclusive)? rustfmt does this for code but not for comments.

BurntSushi (approved):
Thanks!

BurntSushi:
This PR is on crates.io in `csv 1.1.4`.

PR #226, opened 2021-03-06
Remove /examples/data from being included in published cargo package

fschutt:
The /examples directory got published on crates.io, which led to me having 2.48 megabytes of NFL plays in my ~/.cargo directory. While it's certainly interesting to know that O.Daniels managed to do a 14-yard touchdown in 2012, I don't think it's wise to include this info in every copy of rust-csv.

Usually if you want to run examples, you'd just look / clone directly from GitHub, but there's no need to publish example data to crates.io


fschutt:
@BurntSushi can this be merged, please?

PR #231, opened 2021-05-03
doc: A note about internally-tagged enums not supported

zeenix:
Internally-tagged enums are extremely useful and therefore used quite a lot in the wild. It's a bit of a shock to users to learn (the hard way) that they're not supported by `csv`. So it'd be great to at least document this limitation in the docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,6 +145,12 @@ $ cd rust-csv
 $ cargo run --example cookbook-read-serde < examples/data/smallpop.csv
 ```
 
+## Internally-tagged enums
+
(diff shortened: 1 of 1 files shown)


zeenix:
@BurntSushi could you please have a look?

zeenix:
@BurntSushi Given that you explicitly asked me for this PR and it's a simple change that will likely help a lot of people avoid bumping into the serious and unexpected limitation, ignoring it for years, is very disrespectful to say the least. 😔

wiktor-k:
I just found this ticket after trying to get internally tagged enums to work... it'd be really nice if this was merged to at least warn users about the limitations :roll_eyes:

PR #233, opened 2021-05-28
Allow for escaping the delimiter like the python counterpart

### LINKED ISSUE #232: Allow escaping the delimiter

1.1.6

Allow escaping the delimiter like the [python counterpart](https://docs.python.org/3/library/csv.html#csv.Dialect.escapechar), e.g.
```
field1,field2,field3\, with\, commas,field4
```

This is to me, personally, the best format for CSV. No need for quotes or anything else, it's just a delimiter that can be escaped if used within a field. I currently have code in python that writes files like this, and I'm in need of reading them through rust. Of course this might be possible already, but I've not had any success with the `ReaderBuilder` options, also the `escape` function mentions escaped *quotes* only.

#### What is the observed behavior of the code above?

```
CSV error: record 19 (line: 19, byte: 5289): found record with 20 fields, but the previous record has 18 fields
```

I will link a PR to this issue which made it work for me, but the implementation of it is probably sub-optimal.

leaty:
Fixes #232 

As mentioned in the issue, this is probably sub-optimal, and it might even break stuff. In fact this might even be possible already, just hidden in the docs. In any case I would just like to hear some ideas.

Using these options:
```rust
let mut csv = ReaderBuilder::new()
    .delimiter(b',')
    .quote(b'\\')
    .quoting(true)
    .escape(Some(b'\\'))
    .flexible(false)
```

One can now read this properly:
```
field1,field2,field3\, with\, commas,field4
```


csv-core/src/reader.rs
```diff
@@ -1023,7 +1025,9 @@ impl Reader {
                 }
             }
             InQuotedField => {
-                if self.quoting && self.quote == c {
+                if self.delimiter == c {
+                    (InField, NfaInputAction::CopyToOutput)
```
BurntSushi, line 1029: I don't see how this could possibly be right..? This is saying that if the parser is in a quoted field and it sees a delimiter, then it should pop out of the quoted field immediately. But the whole point of quoted fields is to treat the delimiter as just another character. So I don't think this state should be doing any case analysis on the delimiter at all.

leaty:
Thanks for the quick reply. All of what you said makes sense, apologies for the poor (and quite clearly broken code). I just needed this to work quickly and I thought I'd share my current changes with a PR to spur a conversation.

Using your notes I thought of this:
1. Add a new function to the builder which sets e.g. `escapechar: Option<u8>`. This would then be a true escape character (escapes any one character immediately after).
2. Add a new state that is set after e.g. `if self.escapechar = Some(c)` while currently in the `InField` state.
3. Exit this new state on the next character and return back to `InField`.

Would that be something that could be considered for this csv library? I know it's not *standard* csv, but it is technically still *comma-separated values* albeit with an escape character.

Again, apologies for the code.

BurntSushi:
No need to apologize. I'm happy to receive lower effort stuff first: it lets us prove out the idea, and ultimately if we don't move forward with it, then not much has been lost.

Your plan sounds okay I think, although, I wonder if it makes sense to re-use the current `escape` option, and then add a new (disabled by default) boolean option that permits the escape character to apply outside of quoted fields. e.g., `escape_outside_quoted_fields` or something like that.

leaty:
> I wonder if it makes sense to re-use the current `escape` option

Yes, I'm all for reusing that. I also think it's better if it's only enabled when quoting is false *and* the escape character is set. In most if not all cases, you either run with quotes for escaping the value *or* an escape character to escape the actual delimiter. There shouldn't really be a good reason to use both ways at the same time.

From python std docs for escapechar in csv:
> A one-character string used by the writer to escape the delimiter if quoting is set to QUOTE_NONE and the quotechar if doublequote is False. On reading, the escapechar removes any special meaning from the following character. It defaults to None, which disables escaping.

Would that be manageable? The escape option is never used if quoting is disabled correct? If I'm right, then you could re-use the escape character for this, quite freely. Assuming quoting also disables double quote, it might look like this:
```rust
let mut csv = ReaderBuilder::new()
    .quoting(false)
    .escape(Some(b'\\'))
    ...
```

BurntSushi:
Oh I see. Off hand, I believe that's right, so no new config knob is needed. Nice.

leaty:
Sounds good, I'll give it a go this weekend :+1:

leaty:
I must be misinterpreting something, I really thought something like this would work. But it's still interpreting `,` after a `\` as a delimiter and giving me `UnequalLengths`. Any ideas?

leaty:
Solved by moving escape out from under quoting: https://github.com/BurntSushi/rust-csv/blob/91998f68c0c6b3b6ea77eb7307115f51cfba9aae/csv-core/src/reader.rs#L807-L812

However, while this doesn't give an error and partly works, it's removing the escaped `,` E.g.
```
field1,field2,field3\, with\, commas,field4
```
Becomes:
```rust
StringRecord(["field1", "field2", "field3 with commas", "field4"])
```

What am I missing?

BurntSushi:
Not sure. I went and looked for the obvious thing (not using `NfaInputAction::CopyToOutput`), but it looks like you're using that. I'd have to debug it, and I doubt I'll have time to do that any time soon unfortunately.

I would suggest debugging it at the `csv-core` level. There are lots of tests in that crate. Add one there (you'll have to do that anyway), and see if you can track it down.

leaty:
Some gdb later and I felt silly, I forgot to add it back to `const NFA_STATES`, after that the `,` remained as it should. I also had to add it to `StartField` in `transition_nfa`, otherwise it wouldn't play nice with fields that *begin* with an escape sequence e.g. `field1,\,field2,field3`.

One thing I'm unsure about is the `Dfa` struct, I put `in_escape_sequence` there because it felt right, but honestly I'm not entirely sure this is necessary in this case. This is the only reference I could find for the others, and it works without adding it:
https://github.com/BurntSushi/rust-csv/blob/938a3723ac05e5fa8f3790d1065765bac35e5abf/csv-core/src/reader.rs#L685-L689

On another note, *if* this is accepted as a valuable feature, could adding it to the writer be considered as well?

PR #247, opened 2021-09-24
Non-allocating trim method

do4gr:
Hey, thanks for the library. 

When I was using it I saw that a lot of time was spent in the trim method and stumbled over the comment about making it non-allocating. 

I took a stab at it and cargo bench shows a big improvement. Let me know if that is something that you would be interested in , than I can keep working on it a bit more.

Below is the output of the included bench functions for trimming. I think they are showing the best case though since the test file seems to contain no trimmable white-space. 
Also what is confusing to me is that the StringRecord benchmark also benefits from improvements to the ByteRecord trim method, not sure where this is coming from. 

Screen Shot 2021-09-24 at 21 59 36

### DIFF
--- a/src/byte_record.rs
+++ b/src/byte_record.rs
@@ -361,7 +361,7 @@ impl ByteRecord {
     /// record.trim();
     /// assert_eq!(record, vec!["", "foo", "bar", "b a z"]);
     /// ```
-    pub fn trim(&mut self) {
+    pub fn trim_original(&mut self) {
         let length = self.len();
         if length == 0 {
             return;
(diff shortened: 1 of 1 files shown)


do4gr:
I cleaned up the code a bit and added a bench case that actually has to remove whitespace. There the difference is less stark, but still pretty big. 

Screen Shot 2021-10-02 at 12 36 10
Screen Shot 2021-10-02 at 12 37 02

Tristramg:
Thank you for this pull request. This yields a significant performance improvement in our project.

@BurntSushi what would be the next steps to consider this PR?

BurntSushi:
I need to review it. From a quick glance, it looks like there's some clean up required here. For example, there shouldn't be a "trim original" routine in the public API.

CI also obviously needs to be fixed.

Tristramg:
@do4gr hello, thank you at lot for this. Do you plan to work on this PR again? Otherwise I can try to polish what you did in order to get it merged.

do4gr:
I'll remove the trim_original and then we'll see whether the PR run succeed. I think it was just flakey CI.

do4gr:
Apart from that from my side the PR would be done. There is nothing more I was thinking of including.

do4gr:
That's it from my side. Let me know if there is more that needs to be done.

src/byte_record.rs
```diff
@@ -362,18 +362,53 @@ impl ByteRecord {
...
+        if self.0.bounds.ends[length - 1] == 0 {
+            return;
+        };
+
+        let mut read_pos: usize = 0;
+        let mut write_pos: usize = 0;
+        for element_end_idx in &mut self.0.bounds.ends[..length] {
+            // left trim
+            let mut found_only_whitespace = true;
+            while read_pos < *element_end_idx {
```
robryk, line 381: I'd suggest finding the first non-whitespace position first and then using [slice.copy_within](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_within) to copy the rest. That would (a) make this somewhat easier to read (e.g. obviate the need for found_only_whitespace) and (b) be potentially easier to optimize for the compiler (no need to deduce that once the branch with the copy is taken it will be taken until the loop finishes).

robryk (commented):
Thanks for starting this PR.

src/byte_record.rs
do4gr: Thanks for the review. I did not catch the notification, will have a look soon!

PR #255, opened 2021-12-14
deps: Update to itoa 1.

waywardmonkeys:


Keruspe:
Any chance to get a release with this?

complexspaces:
Friendly ping about a possible release if you have the bandwidth, @BurntSushi .

djc:
@RalfJung in https://github.com/redis-rs/redis-rs/issues/594#issuecomment-1216808008:

> It'd be good to get a semver-compatible update that depends on itoa 1.0. This crate is one of the most popular crates that still depends on itoa 0.4, which uses the deprecated `mem::uninitialized` function. On recent nightly builds of the compiler, that function has become a lot slower in order to mitigate the many ways in which it is used incorrectly throughout the ecosystem. We'd like to move even further and add future-compatibility warnings against `mem::uninitialized`, which would show up everywhere that itoa is in the dependency tree.

@BurntSushi ping?

djc:
Ah sorry, I see this was already being discussed in https://github.com/BurntSushi/rust-csv/issues/271.

BurntSushi:
`csv 1.2.0` is now on crates.io.

PR #259, opened 2021-12-31
add GitHub Actions workflow to run tests

striezel:
This can serve as a replacement for the ceased Travis CI builds.

The builds look like this when they are successful: https://github.com/striezel-stash/rust-csv/actions/runs/1641474135

The script part of the GitHub Actions workflow is basically the [`ci/script.sh`](https://github.com/BurntSushi/rust-csv/blob/9f5455bc6a1ffcbfd7ac993dd87e023dd29cb415/ci/script.sh) ported to workflow syntax.

**Unfortunately I had to bump the MSRV build from rustc 1.33.0 to rustc 1.40.0** in the process, because rustc 1.39 and earlier failed to build. I am not sure whether this is due to some error on my part or due to some months without MSRV 1.33.0 builds where some dependencies or Rust features may have slipped into the code. However, the last Travis CI build with rustc 1.33.0 also failed ([see here](https://travis-ci.org/github/BurntSushi/rust-csv/jobs/774578336)), so I assume it is the later one and 1.40.0 is now the actual MSRV of the csv crate.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,50 @@
+name: Rust
+
+on: [ push, pull_request ]
+
+jobs:
+  rust:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        toolchain: [
+          1.40.0,
+          stable,
+          beta,
+          nightly
+        ]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install Rust toolchain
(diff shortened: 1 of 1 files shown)


BurntSushi:
Done in https://github.com/BurntSushi/rust-csv/pull/298

striezel:
> Done in #298

Good. That's even better than the simple "port" of the Travis CI stuff I did here.

PR #269, opened 2022-02-25
Bump to edition-2021 of Rust

PoignardAzur:
This has the merit of using the new Cargo resolver, which means we don't pull serde-derive by default, which means the project compiles faster.

Free performance boost. It's like C++11 all over again :smile:


PoignardAzur:
Ah. Doesn't work =/

The new resolver [was introduced in 1.51](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html), but looks like `rust-csv`'s current MSRV is 1.33.

Ah well. And I was feeling so clever!

PoignardAzur:
Thinking about it, there's no compelling reason to make this change, since this crate is meant to be a dependency for other projects, and in those projects the root workspace is the once deciding which resolver is used.

Closing this PR.

Issue #280, opened 2022-06-01
Performance issue with the redundancy operations


yyzdtccjdtc:
https://github.com/BurntSushi/rust-csv/blob/41c71ed353a71526c52633d854466c1619dacae4/csv-core/src/reader.rs#L669

For this line of code, it will always update the value of `self.line`, even if the value of it does not change. This introduces some redundant operations.

If we change the code to this, we can eliminate these redundant operations.

```
if input[nin] == b'\n'{
     self.line += 1;
}
```

According to my tests with the ema example with this application (https://github.com/greyblake/ta-rs), the average execution time decreased from 2.74s to 2.68s, which is a 1.02x speedup.

Hope this information helps!

BurntSushi:
Interesting. Seems like the compiler should be smarter than that here. But happy to make this change. PRs are welcome.

PR #281, opened 2022-06-01
Optimization for eliminating redundant memory operations

yyzdtccjdtc:
Optimization for eliminating redundant memory operations related to this issue.
https://github.com/BurntSushi/rust-csv/issues/280

### DIFF
--- a/csv-core/src/reader.rs
+++ b/csv-core/src/reader.rs
@@ -666,7 +666,9 @@ impl Reader {
         let mut state = self.dfa_state;
         while nin < input.len() && nout < output.len() && nend < ends.len() {
             let (s, has_out) = self.dfa.get_output(state, input[nin]);
-            self.line += (input[nin] == b'\n') as u64;
+            if input[nin] == b'\n'{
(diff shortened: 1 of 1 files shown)


yyzdtccjdtc:
@BurntSushi Is there are any changes I need to make for this pull request?

csv-core/src/reader.rs
```diff
@@ -666,7 +666,9 @@ impl Reader {
         let mut state = self.dfa_state;
         while nin < input.len() && nout < output.len() && nend < ends.len() {
             let (s, has_out) = self.dfa.get_output(state, input[nin]);
-            self.line += (input[nin] == b'\n') as u64;
+            if input[nin] == b'\n'{
+                self.line += 1;
+            }
```
adam-azarchs, line 671: Do you have benchmarks backing this up?  I'm a little skeptical that the cost of looking up `self.line` in order to increment it by zero is higher than the cost of adding a branch in the code path, at least on most modern CPUs where branches do unpleasant things to instruction pipelining.

PR #290, opened 2022-10-23
Export record deserializer.

ryzhyk:
The library currently exposes deserialization API through the `Reader::deserialize` method, which returns an iterator over deserialized values.

This commit exposes the deserializer type through the public API, allowing users to create their own deserializer instances from `StringRecord`s and `ByteRecord`s.  This enables interoperability with existing code parameterized by deserializer type.  In my specific use case I convert deserializer into a trait object using `erased_serde`, which allows switching between different data formats at runtime, similar to the example here:
https://docs.rs/erased-serde/0.3.23/erased_serde/trait.Deserializer.html

Signed-off-by: Leonid Ryzhyk <ryzhyk@gmail.com>

### DIFF
--- a/src/deserializer.rs
+++ b/src/deserializer.rs
@@ -18,15 +18,24 @@ use crate::string_record::{StringRecord, StringRecordIter};
 
 use self::DeserializeErrorKind as DEK;
 
+/// Create a deserializer for a single CSV record stored as a UTF-8 string.
+pub fn string_record_deserializer<'de, D: Deserialize<'de>>(
+    record: &'de StringRecord,
+    headers: Option<&'de StringRecord>,
+) -> StringRecordDeserializer<'de> {
+    DeRecordWrap(DeStringRecord {
+        it: record.iter().peekable(),
+        headers: headers.map(|r| r.iter()),
+        field: 0,
+    })
+}
+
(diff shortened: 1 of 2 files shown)


BurntSushi:
Could you please say why `StringRecord::deserialize` is insufficient for your use case? https://docs.rs/csv/latest/csv/struct.StringRecord.html#method.deserialize

I can at least say that this patch has basically zero chance of landing as-is. I do not export internal APIs---especially weird and complex ones like this---without careful consideration and API design.

ryzhyk:
The [`erased-serde`](https://crates.io/crates/erased-serde) crate allows building object-safe interfaces around deserialization code.  For instance, this [example](https://docs.rs/erased-serde/0.3.23/erased_serde/trait.Deserializer.html) implements a scenario where the user can dynamically switch between different serialized representations of the same Rust type, e.g., CBOR and JSON.  My use case is similar (except that it comes up in a real application), and I'd like to add `.csv` to the list of supported formats. As it stands today, `rust-csv` doesn't allow this, as it does not export its deserializer type.

BurntSushi:
If you want to see this progress, you're going to need to open a new issue and explain the problem in a bit more detail with code examples. There also need to be API design work and not just "as it does not export its deserializer type."

Issue #291, opened 2022-10-24
feat: support trim in `csv_core`


youngsofun:
We are using  `csv_core` in our project [databend](https://github.com/datafuselabs/databend),  thank you for your great work!

Currently, trim is done outside  `csv_core`, based on the result of `read_field` or `read_record`.
it works well when there is no quote used in the field.
but leading to 2 kinds of bad cases otherwise.

so I suggest to support trim in `csv_core` directly.

# 1. bad cases

[demo of the problem in rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fadee3357e54fd2c0da1129f98aa579f)

(in this issue, use `|` as field delimiter, use `>value<` to mark the parse result)

## case 1:  spaces inside the quotes should not be removed

`|" foo"|` 

- py(skipinitialspaces=True):: `> foo<` 
- rust csv trim::ALL : `|foo|` (bad case)

 reason: trim is performed after csv-core **removed** `"`

## case 2:   quote immediately after whitespace should be removed 

`| "foo"| `

- py(skipinitialspaces=True): `>foo<`
- rust csv trim::ALL : `>"foo"<` (bad case)

reason: `csv_core` only remove `"` **immediately** after delimiter

# 2. how other lib deal with this

## 2.1 python
the behavior of [python](https://docs.python.org/3/library/csv.html#csv.Dialect.skipinitialspace) seems more reasonable, and consistent

>>> whitespace immediately following the delimiter is ignored

so the `whitespace immediately following the delimiter` is skipped first,  then the remaining is parsed normally.

but in this crate,  the order is reversed.
`| " Hello"|`,  `| Hello|`  both result in `> Hello<`, after read_field (can not distinguish here), but only the whitespace in  `| Hello|` should be trimmed

## 2.2 snowflake
snowflake has the same behavior as python:

```
|"Hello world"|    /* loads as */  >hello world<
|" Hello world "|  /* loads as */  > hello world <
| "Hello world" |  /* loads as */  >hello world<
```

see `TRIM_SPACE` option in https://docs.snowflake.com/en/sql-reference/sql/create-file-format.html#optional-parameters

 FYI:  more discuss in   https://github.com/datafuselabs/databend/issues/8416

BurntSushi:
It's hard for me to digest this issue. Could you please provide Rust programs that I can compile and execute, along with input, actual output and desired output?

youngsofun:
[...]

PR #292, opened 2022-10-24
feat: csv_core support trim_left.

youngsofun:
an implementation to fix https://github.com/BurntSushi/rust-csv/issues/291

only implemented `trim_left`: 

1.  more common: python/pandas only support it, see https://github.com/datafuselabs/databend/issues/8416
2. easier to implement with a state machine

### DIFF
--- a/csv-core/src/reader.rs
+++ b/csv-core/src/reader.rs
@@ -2,6 +2,9 @@ use core::fmt;
 
 use crate::Terminator;
 
+const MAX_NUM_WHITESPACES: usize = 6;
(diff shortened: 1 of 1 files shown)


csv-core/src/lib.rs
```diff
@@ -99,6 +99,8 @@ foo,\"bar,baz\"
 #![deny(missing_docs)]
 #![no_std]
 
+extern crate alloc;
```
BurntSushi, line 102: This is a `no_std` crate and I want it to always be a `no_std` crate. We can't use `alloc` here.

youngsofun: fixed

PR #214, opened 2020-11-03
Update URLs, badges, and use SPDX license format

atouchet:
Update URLs, version number listing, and link to latest version of doc pages.

Closes #194, fixes #248.

PR #297, opened 2023-01-17
Use proper SPDX identifier for license

lstrojny:
Tools that parse license identifiers fail with the current identifier of the csv crate. E.g. `cargo-cyclonedx`:

```
[2023-01-17T14:43:55Z ERROR cargo_cyclonedx::generator] Package csv has an invalid license expression, trying lax parsing (Unlicense/MIT): Invalid SPDX expression: invalid character(s)
```

This PR fixes it.


atouchet:
Note: This change is included in #214.

PR #304, opened 2023-03-06
Add IntoInnerError::into_error()

Bravo555:
Currently, for code that doesn't attempt to recover from errors but just reports the error to the caller, it's forced to return the entire `IntoInnerError`, which is very big compared to `io::Error` which is where the actual error is.

This commit adds `IntoInnerError::into_error()` method that consumes `IntoInnerError` and returns contained `io::Error`. This mirrors the API of [`std::io::IntoInnerError`](https://doc.rust-lang.org/stable/std/io/struct.IntoInnerError.html#method.into_error) in Rust stdlib, which provides such method.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -312,6 +312,15 @@ impl<W> IntoInnerError<W> {
         &self.err
     }
 
+    /// Consumes the [`IntoInnerError`] and returns the error which caused the
+    /// call to [`Writer::into_inner`](crate::Writer::into_inner) to fail.
+    ///
+    /// Unlike [`IntoInnerError::error`], this can be used to obtain ownership
+    /// of the underlying error.
+    pub fn into_error(self) -> io::Error {
+        self.err
+    }
+
     /// Returns the underlying writer which generated the error.
     ///
     /// The returned value can be used for error recovery, such as


src/error.rs
```diff
@@ -312,6 +312,13 @@ impl<W> IntoInnerError<W> {
         &self.err
     }
 
+    /// Consumes the [`IntoInnerError`] and returns the error which caused the call to
+    /// [`Writer::into_inner()`] to fail. Unlike `error`, this can be used to
+    /// obtain ownership of the underlying error.
```
BurntSushi, line 317: Please make sure the docs are wrapped to 79 columns, inclusive. (They don't currently appear to be.)

Also, please use `Writer::into_inner` and drop the parens.

And please use `IntoInnerError::error` instead of `error`.

BurntSushi (changes requested):
Thanks! This LGTM with a couple nits.

>  it's forced to return the entire IntoInnerError, which is very big compared to io::Error which is where the actual error is.

src/error.rs
Bravo555: Docs wrapped, and I managed to link to `Writer::into_iter()` correctly this time.

BurntSushi (approved):
Thanks!

BurntSushi:
This PR is in `csv 1.2.1` on crates.io.

PR #306, opened 2023-03-13
Update version number in docs

HalfVoxel:


BurntSushi:
I've re-worded the setup instructions to avoid mentioning a version number.

PR #311, opened 2023-04-10
Use the proper version for serde in Cargo.toml

JohnPeel:
# Description
If you try using `csv` in a project locked to `serde` versions between `1.0.55` and `1.0.66`, `csv` will fail to compile.

# Reason
`csv` uses the `serde_interger128` macro which wasn't introduced until `serde` version 1.0.67.

# Solution
This pull request bumped the `serde` version to 1.0.67.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,11 +25,11 @@ bench = false
 csv-core = { path = "csv-core", version = "0.1.10" }
 itoa = "1"
 ryu = "1"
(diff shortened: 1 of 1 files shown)


JohnPeel:
Actually, jumped the gun. The `bstr` dev-dependency already requires `serde` 1.0.85.
 
 It probably would compile fine.

PR #317, opened 2023-05-03
Improve error reporting of the serde csv serializer.

jstranik:
Make the error messages to show also the field and value that failed to serialize.

Example of an error before the change:
>  Error Parsing Data: CSV deserialize error: record 1 (line: 2, byte: 549): trailing input.

The new improved error message is now displayed as:
> Error Parsing Data: CSV deserialize error: record 1 (line: 2, byte: 549): trailing input. (Field 'trade_timestamp' has a value '2023-04-04T09:46:45.000Z')

Notice the field name and field value that caused problem during the import.


schell:
This would be great. Is there any other way to map an error back to the field that failed to parse?

PR #322, opened 2023-05-31
Add `Writer::get_ref`

threecgreen:
It's useful to be able to access the underlying writer without dropping `Writer`, for example, checking the position of an `std::io::Cursor`.

### DIFF
--- a/src/writer.rs
+++ b/src/writer.rs
@@ -1068,6 +1068,11 @@ impl<W: io::Write> Writer<W> {
         Ok(())
     }
 
+    /// Return a reference to the underlying writer.
+    pub fn get_ref(&self) -> &W {
+        self.wtr.as_ref().unwrap()
+    }
+
     /// Flush the contents of the internal buffer and return the underlying
     /// writer.
     pub fn into_inner(


BurntSushi:
`get_ref` seems okay, but I'm less sure about `get_mut`. It looks like a footgun to me, although a `Reader` does have it.

Does `get_ref` unblock your specific use case? If so, let's just go with that for now.

threecgreen:
Thanks for the quick response!

Yes, `get_ref` unblocks my specific use case.

src/writer.rs
```diff
@@ -1079,6 +1079,11 @@ impl<W: io::Write> Writer<W> {
         }
     }
 
+    /// Return a reference to the underlying writer.
+    pub fn get_ref(&self) -> &W {
+        self.wtr.as_ref().unwrap()
+    }
+
```
BurntSushi, line 1086: Sorry, one more thing. Can you move this above `into_inner` to match the order of methods on `csv::Reader`? Thanks.

BurntSushi (changes requested):

BurntSushi:
This PR is on crates.io in `csv 1.2.2`.

threecgreen:
Thank you!

PR #332, opened 2023-08-05
added NONE Terminator

### LINKED ISSUE #331: Disable line terminator config

csv: 1.2.2
csv-core: 0.1.10

When using `wtr.write_field()` the following code fails since `wtr.write_record(None::<&[u8]>)?;` has not been called as documented. Adding said line not only adds the missing quoting character but additionally adds `\n` to the respective line. Both `escape` and `terminator` are configurable in `csv::WriterBuilder` but is there a way to disable them entirely?

no closing quote
```rust
fn example1() -> Result<(), Box<dyn Error>> {
    let mut wtr = Writer::from_writer(vec![]);
    wtr.write_field("hallo \" world")?;

    let data = String::from_utf8(wtr.into_inner()?)?;
    assert_eq!(data, "\"hallo \"\" world\""); //  `"\"hallo \"\" world"`
    Ok(())
}
```

additional line feed

```rust
fn example2() -> Result<(), Box<dyn Error>> {
    let mut wtr = Writer::from_writer(vec![]);
    wtr.write_field("hallo \" world")?;
    wtr.write_record(None::<&[u8]>)?;

    let data = String::from_utf8(wtr.into_inner()?)?;
    assert_eq!(data, "\"hallo \"\" world\""); // `"\"hallo \"\" world\"\n"`
    Ok(())
}
```

desired behavior

```rust
fn example3() -> Result<(), Box<dyn Error>> {
    let mut wtr = WriterBuilder::new()
        .terminator(None::<&[u8]>)) // or similar ???
        .from_writer(vec![]);

    wtr.write_field("hallo \" world")?;
    wtr.write_record(None::<&[u8]>)?;

    let data = String::from_utf8(wtr.into_inner()?)?;
    assert_eq!(data, "\"hallo \"\" world\""); // \"hallo \"\" world\"
    Ok(())
}
```

#### some background for this request

At the moment I am trying to contribute to the open source tool [vector](https://github.com/vectordotdev/vector). Specifically [#17261](https://github.com/vectordotdev/vector/issues/17261). The tricky part is that the csv crate is only used as a line encoder since the so called [framing](https://vector.dev/docs/reference/configuration/sources/stdin/#framing) is done independently by that tool. 

https://github.com/vectordotdev/vector/blob/4b80c714b68bb9acc2869c48b71848d11954c6aa/lib/codecs/src/encoding/format/csv.rs#L78-L100

Therefore I am curious if there is a way to disable the terminator? If not I would be quite happy if the [Terminator](https://github.com/BurntSushi/rust-csv/blob/master/csv-core/src/lib.rs#L118) enum could add an additional special value implementing this behavior.

In regards to the `escape()` character from what I am seeing the `csv::QuoteStyle::Never` would likely disable all quoting / escaping? This would have been my second request to handle corner cases where such behavior is desired.

https://github.com/BurntSushi/rust-csv/blob/574ae1ff64693b42ae0ce153926d9a0a5d546936/csv-core/src/lib.rs#L174-L175

scMarkus:
Would fix #331

This pull request would enable `rust-csv` to be used as a "csv line writer" with proper quoting but without appending any line terminator. 

Critique on the technical level as well as the semantically one would be highly appreciated so this may become a valuable addition.

### DIFF
--- a/csv-core/src/lib.rs
+++ b/csv-core/src/lib.rs
@@ -120,6 +120,9 @@ pub enum Terminator {
     CRLF,
     /// Parses the byte given as a record terminator.
     Any(u8),
+    /// Disable line terminator
+    /// This is useful when writing single line field by field with proper qoting
+    NONE,
     /// Hints that destructuring should not be exhaustive.
     ///
     /// This enum may grow additional variants, so this makes sure clients
--- a/csv-core/src/writer.rs
+++ b/csv-core/src/writer.rs
@@ -54,6 +54,9 @@ impl WriterBuilder {
             Any(b) => {
                 wtr.requires_quotes[b as usize] = true;
             }
+            NONE => {
+                wtr.requires_quotes[usize::MIN] = true;
+            }
             _ => unreachable!(),
         }
         wtr
@@ -375,6 +378,7 @@ impl Writer {
         let (res, o) = match self.term {
             Terminator::CRLF => write_pessimistic(&[b'\r', b'\n'], output),
             Terminator::Any(b) => write_pessimistic(&[b], output),
+            Terminator::NONE => write_pessimistic(&[], output),
             _ => unreachable!(),
         };
         if o == 0 {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -221,6 +221,9 @@ pub enum Terminator {
     CRLF,
     /// Parses the byte given as a record terminator.
     Any(u8),
+    /// Disable line terminator
+    /// This is useful when writing single line field by field with proper qoting
+    NONE,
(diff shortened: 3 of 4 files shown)


pront:
Hello @BurntSushi, can we please get a review on this PR?

BurntSushi:
PR review latency for me can be quite large, sorry.

The functionality being added here doesn't make sense to me. See: https://github.com/BurntSushi/rust-csv/issues/331#issuecomment-1674911636

BurntSushi:
See https://github.com/BurntSushi/rust-csv/issues/331#issuecomment-1679302469

Issue #349, opened 2024-02-01
How to serialize to a byte buffer


huntc:
Is there a way to serialize to a byte buffer and then write those buffered bytes out periodically? I'm establishing my writer like this:

```rust
    let mut buffer = csv::Writer::from_writer(BufWriter::new(Vec::new()));
```

That is then passed into a map closure on a stream. I want my stream map closure to serialize the fields passed to it and then return the bytes not previously returned.

huntc:
Adding to this, I think my issue is that I can't get a mutable ref to the underlying buffer of the CSV writer. If I could, then I could clear the underlying vector.

BurntSushi:
The csv writer owns the buffer. It wouldn't really make sense to expose a mutable ref to it. There is into_inner though.

To be honest, I don't quite understand what you're trying to do without code. My instinct is to suggest that you provide a Write implementation that does the periodic writing.

huntc:
Here's some code:

```rust
    let mut buffer = csv::Writer::from_writer(BytesMut::new().writer());
    let event_stream = ReceiverStream::new(gateway_events_receiver)
        .map_while(|m| m)
        .map(move |(_entity_id, event)| {
            let _ = buffer.serialize(event);
            let _ = buffer.flush();
            let bytes = buffer.get_ref().get_ref().to_vec();
            Ok::<Vec<u8>, Infallible>(bytes)
        })
```

So, I establish a CSV writer which is then moved into my Stream's mapping of events. I want the mapper to write out the bytes it has and then reset the buffer. At the moment, I can't reset the buffer so it will just grow.

BurntSushi:
Why can't you use into_inner, clear the buffer and then rebuild the writer?

huntc:
> Why can't you use into_inner, clear the buffer and then rebuild the writer?

So, rebuild the writer for every line in my CSV file? I could have thousands of lines; might that be inefficient?

Also, doesn't the CSV write hold state?

I've just added a PR and modified my code to use it. I think a mutable ref here is fairly consistent with other libraries, including std.

```rust
    let mut buffer = csv::Writer::from_writer(BytesMut::new().writer());
    let event_stream = ReceiverStream::new(gateway_events_receiver)
        .map_while(|m| m)
        .map(move |(_entity_id, event)| {
            let _ = buffer.serialize(event);
            let _ = buffer.flush();
            let bytes_mut = buffer.get_mut().get_mut();
            let bytes = bytes_mut.split().to_vec();
[...]

PR #350, opened 2024-02-01
Mutable ref

huntc:
Provides a mutable ref method of the underlying buffer so that it can be operated on. This is particularly useful when taking the bytes of the underlying buffer and writing them out in a stream. Upon writing them out, the underlying buffer can be cleared.

See https://github.com/BurntSushi/rust-csv/issues/349 for more background.


huntc:
Closing as rust-csv is more generally unsuitable for use with async streams.

PR #383, opened 2025-02-12
Fix Reader::reset not resetting output_pos

### LINKED ISSUE #382: Reader::reset doesn't reset ends
### Bug

A little hard to explain in words (might be a little late here), but here is a [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eb85fe146010fde6cd78bc4a6addf580) to reproduce.

`Reader::reset` doesn't seem to reset record field ends. When used on a new slice after partially processing an unterminated row, the ends are returned from the previous slice, not the new one.

levkk:
Fix #382

### DIFF
--- a/csv-core/src/reader.rs
+++ b/csv-core/src/reader.rs
@@ -487,6 +487,7 @@ impl Reader {
         self.nfa_state = NfaState::StartRecord;
         self.line = 1;
         self.has_read = false;
+        self.output_pos = 0;
     }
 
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
Thanks!

BurntSushi:
This PR is on crates.io in `csv-core 0.1.12`.

PR #392, opened 2025-03-18
Address Clippy lints

thaliaarchi:
Address several small Clippy suggestions.

### DIFF
--- a/src/byte_record.rs
+++ b/src/byte_record.rs
@@ -698,7 +698,7 @@ impl Bounds {
     /// If there are no fields, this returns `0`.
     #[inline]
     fn end(&self) -> usize {
-        self.ends().last().map(|&i| i).unwrap_or(0)
+        self.ends().last().copied().unwrap_or(0)
     }
 
     /// Returns the number of fields in these bounds.
--- a/src/serializer.rs
+++ b/src/serializer.rs
@@ -1,15 +1,12 @@
 use std::{fmt, io, mem};
 
-use {
-    itoa, ryu,
-    serde::{
-        ser::{
-            Error as SerdeError, Serialize, SerializeMap, SerializeSeq,
-            SerializeStruct, SerializeStructVariant, SerializeTuple,
-            SerializeTupleStruct, SerializeTupleVariant, Serializer,
-        },
-        serde_if_integer128,
+use serde::{
+    ser::{
+        Error as SerdeError, Serialize, SerializeMap, SerializeSeq,
+        SerializeStruct, SerializeStructVariant, SerializeTuple,
+        SerializeTupleStruct, SerializeTupleVariant, Serializer,
     },
+    serde_if_integer128,
 };
 
 use crate::{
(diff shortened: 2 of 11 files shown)


BurntSushi:
Please break this into 3 separate PRs.

thaliaarchi:
I've split out https://github.com/BurntSushi/rust-csv/pull/393 and https://github.com/BurntSushi/rust-csv/pull/394.

src/byte_record.rs
```diff
@@ -50,7 +50,7 @@ impl<T: AsRef<[u8]>> PartialEq<Vec<T>> for ByteRecord {
     }
 }
 
-impl<'a, T: AsRef<[u8]>> PartialEq<Vec<T>> for &'a ByteRecord {
+impl<T: AsRef<[u8]>> PartialEq<Vec<T>> for &ByteRecord {
```
BurntSushi, line 53: I prefer the explicit lifetime here.

src/byte_record.rs
```diff
@@ -62,7 +62,7 @@ impl<T: AsRef<[u8]>> PartialEq<[T]> for ByteRecord {
     }
 }
 
-impl<'a, T: AsRef<[u8]>> PartialEq<[T]> for &'a ByteRecord {
+impl<T: AsRef<[u8]>> PartialEq<[T]> for &ByteRecord {
```
BurntSushi, line 65: I prefer the explicit lifetime here.

src/serializer.rs
```diff
@@ -30,7 +27,7 @@ struct SeRecord<'w, W: 'w + io::Write> {
     wtr: &'w mut Writer<W>,
 }
 
-impl<'a, 'w, W: io::Write> Serializer for &'a mut SeRecord<'w, W> {
+impl<W: io::Write> Serializer for &mut SeRecord<'_, W> {
```
BurntSushi, line 30: I prefer the explicit lifetime here.

BurntSushi: And similarly for the things below.

src/tutorial.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::needless_doctest_main)]
```
BurntSushi, line 1: Please remove.

src/writer.rs
```diff
@@ -1103,6 +1103,7 @@ impl<W: io::Write> Writer<W> {
 
     /// Flush the contents of the internal buffer and return the underlying
     /// writer.
+    #[allow(clippy::result_large_err)]
```
BurntSushi, line 1106: Please remove.

BurntSushi (changes requested):
A lot of these are because of MSRV bumps, there code can be marginally simplified because of new Rust features/APIs. Those are generally fine.

But at least some of these are preference things that I don't agree with. This is why I don't have Clippy running on my repos, because there's too much that Clippy does by default that I don't agree with. I've long wanted to sit down and find the subset of lints I _do_ agree with, but this is a lot of work to do. And then applying it to every other repo I have is just a ton of busy work. And honestly, most of the Clippy lints flagged in my code is more stylistic than it is catching bugs.

thaliaarchi:
Sorry for all the nits. I probably should have kept this to just `#[non_exhaustive]`.

thaliaarchi:
Hey, friendly ping :). Since you already did a review and it's somewhat fresh, I just want to check in.

It seems like we've filtered out the objectionable bits, so it might be ready. I know it's an unimportant change, but I'd like to resolve this PR or even, since you don't usually use Clippy, just close it.

BurntSushi (approved):
LGTM. Thanks!

Issue #325, opened 2023-06-16
Use `#[non_exhaustive]` tag instead of manually `__Nonexhaustive` variant


stackinspector:
There is already a feature for this and we can prefer the "native" way.

PR #393, opened 2025-03-18
api: Use `#[non_exhaustive]` instead of `#[doc(hidden)]` variant

thaliaarchi:
This was introduced in Rust 1.40, which is below the MSRV of 1.73.

It technically breaks any programs which use the undocumented `__Nonexhaustive` variants.

Fixes https://github.com/BurntSushi/rust-csv/issues/325


BurntSushi (approved):
Thanks!

It is possible I'll have to revert this because of practical breakage if it's widespread. But `__Nonexhaustive` is absolutely _not_ part of the public API, so anyone using it has explicitly opted out of semver guarantees.

PR #394, opened 2025-03-18
api: impl `Default` for `Position`

thaliaarchi:
This was suggested by a Clippy lint.

### DIFF
--- a/src/byte_record.rs
+++ b/src/byte_record.rs
@@ -642,6 +642,13 @@ impl Position {
     }
 }
 
+impl Default for Position {
+    #[inline]
+    fn default() -> Self {
+        Position::new()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


BurntSushi (commented):
I'm not so sure about this one. I'm not sure `Position` really has a "default" value. Arguably, `Position::new` is somewhat poorly named, and would be better named as `start` or something.

I would rather there be a bit of a speed bump here over folks blindly deriving `Default` on `Position`.

To be clear, I don't feel strongly about this. I could bow to practical use cases that want a `Default` impl, but a Clippy lint isn't one of them.

thaliaarchi:
I don't mind. Sounds like you're leaning on no, so feel free to close.

PR #400, opened 2025-08-17
test: fix broken test after rust-lang/rust#115746

paolobarbolini:
https://github.com/rust-lang/rust/pull/115746 made the thread id get printed in panic messages, making the `tutorial_error_01_errored` test fail because it matched on the old panic message format.


BurntSushi (approved):
Thanks!

PR #414, opened 2025-10-29
Added `#[no_std]` marker to csv core

LucaCappelletti94:
As per title, this simple PR adds the `#[no_std]` marker to the `csv-core` crate. This way, if someone happens to add any std stuff in the crate, they will immediately get an error message about it.

### DIFF
--- a/csv-core/src/lib.rs
+++ b/csv-core/src/lib.rs
@@ -17,6 +17,7 @@ the `Writer` API provides a CSV writer.
 This example shows how to count the number of fields and records in CSV data.
 
 ```
+#[no_std]
 use csv_core::{Reader, ReadFieldResult};
 
 let data = "


csv-core/src/lib.rs
```diff
@@ -17,6 +17,7 @@ the `Writer` API provides a CSV writer.
 This example shows how to count the number of fields and records in CSV data.
 
 ```
+#[no_std]
```
Tahaa-Dev, line 20: This should be the inner crate attribute `#![no_std]`, not `#[no_std]`. Put `#![no_std]` as the very first line inside the doctest so the example crate is compiled without the standard library.

#### Why?

- `#![no_std]` is an inner crate attribute and is how you opt a crate (or doctest crate) out of the standard library.

- `#[no_std]` is an outer attribute that would apply to the next item and is not the correct way to disable std,  it will either be ignored or cause an unknown-attribute error.

Tahaa-Dev (changes requested):
The line you added is not the correct way to disable std, you should add `#![no_std]` instead.

